PR #5, opened 2015-06-04
add License, add CMS, dev tools, image galleries, link sharing platforms, system monitors

nodiscc:
See commit messages.


daftspunk:
Thanks for the shout out @nodiscc! :peace_symbol:

PR #18, opened 2015-06-28
add more software, add password managers section

nodiscc:


nodiscc:
@Kickball this now merges properly.
I did not sort added lines, can you do it after the merge?
Thanks

PR #27, opened 2015-07-14
add OpenFire, fix elgg, add new category "webmail interfaces"

TwizzyDizzy:


nodiscc:
Shouldn't webmails go to the `Communication systems` section? We can split this section later if the list grows too large

TwizzyDizzy:
Yeah, sure, why not. If there will be more we just split it up again.

PR #33, opened 2015-07-27
Suggest MPD, Subsonic, Ampache

remyhonig:


nodiscc:
Check https://github.com/Kickball/awesome-selfhosted/pull/34

PR #36, opened 2015-07-27
Added most Atlassian products

Mika56:
Hi,

As a developer, I cannot recommend Atlassian products enough. They are very well done softwares that I use everyday in my job.
Although not free, the first pricing tier is affordable ($10 per year for 10 users) and can be self-hosted

### DIFF
--- a/README.md
+++ b/README.md
@@ -445,5 +445,6 @@ The list of contributors can be found below
      1  Igor Antun <IgorAntun@users.noreply.github.com>
      1  Jeremiah Marks <jeremiah@jlmarks.org>
      1  Joshua Hamilton <joshua.hamilton@fabricut.com>
+     1  Mikael Peigney <Mika56@users.noreply.github.com>
      1  Remy Honig <remyhonig@users.noreply.github.com>
      1  stevesbrain <stevesbrain@users.noreply.github.com>
--- a/README.md
+++ b/README.md
@@ -84,6 +84,7 @@ Table of Contents
   * [Gaskit](https://github.com/bkeepers/gaskit) - a git-backed issue tracker. It uses a branch in your local git database to store the tickets. `GPL`
   * [HabitRPG](http://habitrpg.com/) - A habit tracker app which treats your goals like a Role Playing Game. ([Source code](https://github.com/lefnire/habitrpg))
   * [Kanboard](http://kanboard.net/) - A simple and open source visual task board `AGPLv3`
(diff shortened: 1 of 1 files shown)


nodiscc:
Thanks; Few remarks about added software

 * Hipchat:
  * I'd better link to https://www.hipchat.com/server, which is the self-hostable software page itself
  * Uses a third party service (_our cloud-based video servers_) for video streams (video can be disabled) (https://confluence.atlassian.com/display/HC/HipChat+Server+FAQs)

Other products have both a "Cloud" based version, and a "Server" (self-hostable) version, for a price.

@Mika56 What happens when the atlassian subscription ends, and you're using a self-hosted edition? Does the product lock itself?

Mika56:
Nice catch about video streams being sent to their cloud servers, I didn't know that!
Do you want me to submit a new PR?

About subscriptions, this is true for all products, be here's the link for JIRA: https://www.atlassian.com/licensing/jira#serverlicenses-9
Short answer, when your subscription ends, you will not get access to newer versions, but your can still use the software (*Perpetual JIRA use*).

You can, indeed, use their cloud solutions, be we're here to speak self-hosting :)
Anyway, you can get access to their source code if you go over the 10 users price, but that licence does not come with all rights (https://www.atlassian.com/end-user-agreement/ section 6.3 *Your Modifications*)

nodiscc:
> you will not get access to newer versions, but your can still use the software (Perpetual JIRA use).

Looks fair to me.

> Nice catch about video streams being sent to their cloud servers, I didn't know that! Do you want me to submit a new PR?

What should we do?

 * Add a footnote/special symbol for software that sends data to third party services?
 * Leave it as is, expect users to investigate limitations since it's marked as proprietary?
 * Remove it if we decide use of third party servers/backends is a criteria for removal.

I need some time to think about it, but I'd be ok for removal of these services (privacy implications, business implications) if we all agree. Ping @Kickball

Mika56:
I'd say every software is different. In our case, there is indeed traffic sent to cloud servers, but on a functionality that can be disabled. We could add something like
`Privacy issues [1]`
With a footnote:
> [1]: HipChat uses their cloud services for 1:1 video conversations. See [One-to-One Video Chat](https://confluence.atlassian.com/display/HC/One-to-One+Video+Chat)

Kickball:
I think that the best way to handle this would be the footnote as @Mika56 suggested in the above comment.

If we could add the Atlassian products with footnotes for Privacy Issues as needed that would be great.

PR #41, opened 2015-07-27
Add Jekyll static site generator

erimar77:


jalcine:
Yeah, I was surprised this wasn't on the list!

nodiscc:
I don't think we should start referencing static site generators. These websites https://staticsitegenerators.net/ and https://www.staticgen.com/ do a much better job than us at listing, sorting and comparing them.

A subsection for CMS like this would be great:

## Static Site Generators

See https://staticsitegenerators.net/ and  https://www.staticgen.com/ for detailed listings

jalcine:
Woah, didn't notice about these sites :+1:! I'm more of favor of that now than adding the Jekyll entry.

nodiscc:
@erimar77 can you update your patch? Otherwise I'll do it tomorrow

PR #43, opened 2015-07-27
Add PasswordStore to Password Managers

jalcine:
Adding in a Git-backed, GPG-encrypted, CLI-driven (with clients for
other devices and systems) password manager.


nodiscc:
@jalcine I don't think pass qualifies for this list. It does not have a client/server model, does not allow password management for multiple machines over a network, etc... You can still run it on the command line through SSH, which is true for many other software.

This list is about locally hosted software used to replace [SaaS](https://en.wikipedia.org/wiki/Software_as_a_service)

jalcine:
I guess so, re: the SaaS stance. I figured it'd be a useful add for those who wanted to use their own instance of a password manager.

PR #49, opened 2015-07-28
Fix some markdown

nyanpasu:
- Fix link to file sharing and synchronization
- Fix indentations of mind mapping


nodiscc:
Fixed in another commit, thank you.

PR #51, opened 2015-07-28
add tastebin

jungle-boogie:
addition of an updated version of hastebin.
sorting of couple items.
updated license on couple projects.


jungle-boogie:
Is that better?

jungle-boogie:
I think it's been corrected now.

Issue #65, opened 2015-07-29
Static sites generators section is gone


nodiscc:
Content added at https://github.com/Kickball/awesome-selfhosted/commit/d206871605cb3d937ee9e944f4510e24c1f4a2b5 can not be found in the latest master.

Edit: can't find in the commit history where it has been removed. Can someone confirm?
Edit: found https://github.com/Kickball/awesome-selfhosted/commit/1f04a09d795fefde69f5462e7f790d9cabc95fce?short_path=04c6e90#diff-04c6e90faac2675aa89e2176d2eec7d8

PR #66, opened 2015-07-29
Added Nikola and Octopress

cron410:


nodiscc:
@cron410 There used to be a section for static site generators but it's been removed by accident (see https://github.com/Kickball/awesome-selfhosted/issues/65). Basically we point to https://staticsitegenerators.net/ and https://www.staticgen.com/ which are very well designed and complete, instead of listing 10³ static site generators.

I see both these projects are already listed there.

Can we close this PR?

cron410:
Sounds good, I dont know how to close it.

PR #78, opened 2015-07-31
mention jitsi in custom

jungle-boogie:
addition of jitsi meet and their video bridge.


nodiscc:
Thank you. Ok for Jitsi meet, however this is a client for Videobridge; can the Videobridge solution be self-hosted or does Jitsi Meet have to use jit.si servers?

From https://jitsi.org/Projects/JitsiVideobridge it's a mix of [libjitsi](https://jitsi.org/Projects/LibJitsi) and [this XMPP XEP](https://xmpp.org/extensions/xep-0340.html) but the instructions on setting up a server are unclear.

jungle-boogie:
Hello,

The Jitsi video bridge can be used with or without jitsi meet. If you want to use meet, you need the video bridge. If you only want to use the video bridge, you can use the Jitsi Client and an xmpp server for video conferences.

Jitsi meet can be ran by anyone using the video bridge, xmpp server and a few other components. 

Is this better?

Thanks

nodiscc:
I didn't see https://download.jitsi.org/jitsi-videobridge/ ... facepalm.
Thanks

PR #87, opened 2015-07-31
Added Convos to IRC clients

marcusramberg:


Kickball:
Do we need to link to the artistic license?

nodiscc:
It would be better yes, could be opened as a separate issue?

Kickball:
Yep will do :)

PR #98, opened 2015-08-01
Zimbra collaboration

motiko:
Added Zimbra collaboration to Email category.
Free email server (exchange like) hosted solution


Kickball:
If you could remove the spaces between the [text] and (url) then I'm happy to merge this.

@nodiscc what do you think?

motiko:
Sure

Kickball:
On second thought would this be classified as Groupware or Email? It does provide other options and might better fit into Groupware.

What do you think @nodiscc?

motiko:
I used it back when it was primarily a mail server . I guess the focus is still on that side.
But surely enough it can fit into both categories.

Kickball:
Just waiting for @nodiscc opinion on which sections we should add this to (groupware/email).

nodiscc:
Hi everyone, Groupware seems appropriate.

PR #125, opened 2015-08-03
Remove dead link

mertinop:
Pinry Demo


nodiscc:
Thank you for noticing @mertinop

PR #132, opened 2015-08-03
Added FilePizza.

erenhatirnaz:
Added FilePizza to File Sharing and Synchronization.


nodiscc:
Thank you, moved to https://github.com/Kickball/awesome-selfhosted/pull/133

PR #135, opened 2015-08-03
Add OSIAM (osiam.org) to the list.

sputnik27:


Kickball:
Thanks for the contribution!

PR #160, opened 2015-08-04
Add Nimbus

ethanal:
A self-hosted file sharer and URL shortener.


nodiscc:
From the project description

> The files are stored in Amazon S3, so you must have an AWS account.

Unless this is optional and you can also host files on a server of your own, this can't be added unfortunately. @ethanal can you confirm?

ethanal:
Nimbus now works without S3 - setting an env var switches it over to file system storage.

nodiscc:
> If you are not using S3, set the FS_STORAGE variable to TRUE.

This looks right.

PR #170, opened 2015-08-05
Add HuBoard (Bug and Task Trackers)

ChrisMissal:


ChrisMissal:
It connects to GitHub via the API and nothing else is required to use it. Does this work?

> Instant project management for your GitHub issues (Connects directly GitHub API)

nodiscc:
Yes this works, I wish you could add `⚠` next to the title and

`⚠ Depends on a third party network service`

in the licenses list (first item). Is this ok?

ChrisMissal:
@nodiscc sounds fine, I amended the commit

PR #204, opened 2015-08-13
Fix Table of Contents links

CraigyDavi:
This fixes links in the table of contents using the & symbol.


nodiscc:
Thanks @CraigyDavi I think my last merge borked the table of contents

PR #208, opened 2015-08-15
Added PleX

victorheld:
Added PleX


nodiscc:
Already in non-free.md: https://github.com/Kickball/awesome-selfhosted/blob/master/non-free.md

PR #210, opened 2015-08-15
Add wger

rolandgeider:
Added wger to the self-hosting solutions because I couldn't find a better fitting category.


nodiscc:
I think it would better fit in the Analytics section

Kickball:
Do we think [reportr](http://www.reportr.io/) ([Source code](https://github.com/Reportr/dashboard)) would fit into whatever category these fit into?

nodiscc:
Yes, reportr should also be moved to analytics... unless we find a better naming/new section.

rolandgeider:
I just moved wger to the analytics section. I can also move reportr there, but I didn't know how specific you want your pull requests

nodiscc:
thanks @rolandgeider I moved wger, reportr and habitica to a new section "Personal Dashboards". Is this ok?

rolandgeider:
@nodiscc Yes, that makes more sense! (at least for wger, where I would say it's only 30% analytics)

PR #213, opened 2015-08-17
Add Mailcow (Formelly Fufix)

joubertredrat:
Mail server suite based on Dovecot, Postfix and other open source software, that provides a modern Web UI for administration.


nodiscc:
@joubertredrat @Kickball this was already in https://github.com/Kickball/awesome-selfhosted#email , so this is a duplicate. Should we also move Modoboa and Mail-in-a-box to the `Mail` section?

Kickball:
@nodiscc I think if they provide more than just email then they should be in the Self-hosting Solutions or Groupware sections.

nodiscc:
@Kickball ok, then let's remove the mailcow entry in the `Mail` section. We can also add a link at the top of the section:

For preconfigured mail server solutions, see [Self-hosting Solutions](#self-hosting-solutions)

Kickball:
That seems like the best way of doing it.

PR #221, opened 2015-08-18
Improve diaspora* entry

jaywink:
Fix name branding, add link to (random) community server for demo, add source code link, add main programming language.


nodiscc:
Thanks again.

PR #227, opened 2015-08-23
added RetroShare

dimqua:


Kickball:
I'm being picky here, but could you add a space between the GPL and C++ please, then we should be good to merge.

nodiscc:
@dimqua @Kickball I'd prefer if this was _only_ in _Communication Systems_ (no duplicates)

Issue #127, opened 2015-08-03
Reduce overlap with awesome-sysadmin


nodiscc:
Followup to https://github.com/Kickball/awesome-selfhosted/pull/81; our list currently overlaps with @n1trux's https://github.com/n1trux/awesome-sysadmin in many areas. To prevent overlap, avoid NIH syndrome and decrease maintenance on the 2 lists, I suggest merging some of our lists sections as follows:

Import and remove from awesome-sysadmin:
- [x] copy https://github.com/n1trux/awesome-sysadmin#project-management
  - [x] remove section from awesome-sysadmin
- [x] copy https://github.com/n1trux/awesome-sysadmin#statistics
  - [x] remove sectrion from awesome-sysadmin https://github.com/n1trux/awesome-sysadmin/pull/123
- [x] copy https://github.com/n1trux/awesome-sysadmin#ticketing-systems
  - [x] remove section from awesome-sysadmin https://github.com/n1trux/awesome-sysadmin/pull/122
- [x] copy https://github.com/n1trux/awesome-sysadmin#xmpp
  - [x] remove section from awesome-sysadmin - https://github.com/n1trux/awesome-sysadmin/pull/138
- [x] copy https://github.com/n1trux/awesome-sysadmin#webmail
  - [x] remove section from awesome-sysadmin
- [x] copy https://github.com/n1trux/awesome-sysadmin#wiki-software
  - [ ] split file storage and synchronization section in subsections
  - [x] remove section from awesome-sysadmin
- [x] copy https://github.com/n1trux/awesome-sysadmin#cloud-storage
  - [x] remove section from awesome-sysadmin
- [x] ~~copy https://github.com/n1trux/awesome-sysadmin#code-review~~
  - [x] ~~remove section from awesome-sysadmin~~
- [x] copy https://github.com/n1trux/awesome-sysadmin#collaborative-software
  - [x] remove section from awesome-sysadmin https://github.com/n1trux/awesome-sysadmin/pull/301
- [x] copy https://github.com/n1trux/awesome-sysadmin#continuous-integration--continuous-deployment to awesom-sysadmin https://github.com/n1trux/awesome-sysadmin/pull/302
  - [x] remove section from awesome-selfhosted https://github.com/Kickball/awesome-selfhosted/pull/1646

---

Delegate and link to awesome-sysadmin:
- [x] ~~link to https://github.com/n1trux/awesome-sysadmin#control-panels~~
[...]

PR #232, opened 2015-08-27
Add Zabbix to list of monitoring systems.

ekakela:


nodiscc:
Hi, we are currently moving the monitoring section to https://github.com/n1trux/awesome-sysadmin#monitoring (see https://github.com/Kickball/awesome-selfhosted/issues/127). Zabbix is already in that list.

Thank you for the patch anyway, feel free to add other software.

PR #235, opened 2015-08-30
TagSpaces 

unnikked:
TagSpaces is an open source personal data manager. It helps you organize files with tags on every platform.


nodiscc:
It looks like this is not a server side program and runs purely locally on desktop/laptop machines.

> No Backend ➜ No Login ➜ No Cloud

> TagSpaces is running completely offline on your computer, smartphone or tablet and does not require internet connection or online registration. You can still use platforms like ownCloud, Dropbox or Bittorrent Sync in order to sync your files between devices.

So I don't think it can be included.

unnikked:
Given that from repo description

> This is a list of Free Software network services and web applications which can be hosted locally.

and from 

> You can still use platforms like ownCloud, Dropbox or Bittorrent Sync in order to sync your files between devices.

I think it can be included as a support tool

nodiscc:
"sync your files between devices" means synchronizing your data directories between devices. A local tagspaces install is required to browse them with the program. It only runs locally/client side and is actually a file manager. I used it in the past and thought it was good (also try [TMSU](http://tmsu.org/)), but still don't think it belongs here. 

@Kickball @unnikked it it ok to close this issue?

PR #340, opened 2015-11-09
Added RSS Monster

veloxy:


nodiscc:
thanks!

Kickball:
#408 adds RSS Monster.

PR #358, opened 2015-12-02
Add Factor.io to the Continuous Integration section

akhyrul:
From the homepage:

>  "It's like IFTTT (if-this-then-that) for Dev and DevOps"
> Factor.io is a simple way to automate developer workflows by defining them programmatically using the Factor.io syntax and running them using the Factor.io Server.

Still not sure about the license. Moreover I cannot find the source code anywhere other than Ruby Gem repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -678,6 +678,7 @@ See **[Ticketing](#ticketing)**
 
 ### Continuous Integration
 
+  * [Factor.io](https://factor.io/) - It's like IFTTT (if-this-then-that) for Dev and DevOps ([Source code](https://github.com/factor-io/factors)) `MIT` `Ruby`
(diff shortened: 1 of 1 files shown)


nclements:
Looks like the source code is here: https://github.com/factor-io/factor

And license is MIT.

nodiscc:
This should be added to https://github.com/Kickball/awesome-selfhosted#continuous-integration instead of automation

akhyrul:
@nclements @nodiscc thanks! 
should I amend the commit or just create a new one?

nodiscc:
@akhyrul as you wish, doesn't matter.

akhyrul:
@nodiscc done, PR updated

PR #365, opened 2015-12-11
Added IRCAnywhere

pengc99:
Added IRCAnywhere - an open-source self-hosted alternative to IRCCloud


Kickball:
In the source section you open a bracket but never close it.

Once that is fixed this should be good to merge.

pengc99:
Fixed!

PR #370, opened 2015-12-14
Add edx

NamPNQ:


Kickball:
I believe the proper format for this should be:

[edX](https://edx.org) - The Open edX platform is open-source code that powers edx.org. ([Source code](https://github.com/edx/)) `AGPLv3` `Python`

But I'd like to see what @nodiscc thinks.

NamPNQ:
@Kickball Tks for feedback, updated format

nodiscc:
The correct format is

    `AGPLv3` `Python`

Kickball:
@NamPNQ any progress on the formatting fixes?

Kickball:
Moved to #409 because of inactivity.

PR #374, opened 2015-12-16
Remove broken link

brioscaibriste:
reportr.io seems to have expired DNS


nodiscc:
Please don't remove the entry and link to https://github.com/Reportr/dashboard instead

Kickball:
Please move Reportr to be in alphabetical order in the list.

PR #379, opened 2015-12-18
added jingo

mariusv:
Added jingo NodeJS wiki based on git


nodiscc:
Demo should be between parentheses.

Thanks

mariusv:
Demo is now between parentheses also added the source code

Kickball:
Please move it to be in alphabetical order within the section, then it should be good to merge.

mariusv:
Done :+1:

PR #393, opened 2016-01-07
Add Pass

fcharlier:


nodiscc:
Hi, pass has been rejected in the past due to reasons mentioned here: https://github.com/Kickball/awesome-selfhosted/pull/43. Thanks for your patch and feel free to send more, or check the existing issues. Merci encore.

PR #399, opened 2016-01-14
Add TournamentMango

Kickball:
TournamentMango is a games tournament organisation system. Currently it is in the games section, however we could look at creating an entirely new section or sub-section for this project.

What do you think @nodiscc?


nodiscc:
I think the `Games` category is fine. Nice find.

PR #400, opened 2016-01-14
Add Loomio (Forums/Custom)

Kickball:
Loomio is a unique project as it is built around the decision making process and does not fit perfectly into any of our current categories.

I would say that it best fits into either Communication systems/Custom or Communication systems/Social Networks and Forums.

What do you think @nodiscc?


nodiscc:
There are projects that do not fit in a single category. This problem will persist unless we move to a tag-based organization (which is hard to do properly in a simple text file. I think the category you proposed (Social Networking/Forums) is ok. Another possibility would be `Project Management`, but it's currently related to Software Development (subcategory) so not a very good fit.

Feel free to merge.

Kickball:
Thanks for the feedback, I'll give tag based systems a look.

PR #408, opened 2016-01-18
Feedreader Updates

Kickball:
* Add RSS Monster.
* Remove Tiny Tiny RSS Mobile (Last Commit Aug 2014, multiple reports of being broken)
* Added Licenses and Programming Languages
* Removed Broken Links

### DIFF
--- a/README.md
+++ b/README.md
@@ -382,7 +382,7 @@ _Recipe management_
   * [Selfoss](http://selfoss.aditu.de/) - The new multipurpose rss reader, live stream, mashup, aggregation web application ([Source code](https://github.com/SSilence/selfoss)) `AGPLv3` `PHP`
   * [Stringer](https://github.com/swanson/stringer) - A work-in-progress self-hosted, anti-social RSS reader. `MIT` `Ruby`
   * [Tiny Tiny RSS](https://tt-rss.org/gitlab/fox/tt-rss) - Open source web-based news feed (RSS/Atom) reader and aggregator. ([Public instance](http://framanews.org/)) `GPLv3` `PHP`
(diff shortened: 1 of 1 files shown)


nodiscc:
Good job.

About tt-rss-mobile, can you point to the reports asserting it's broken? Last time I used it it was fine (and faster than the native tt-rss UI). I thought the bug reports were mainly feature requests.

Kickball:
I was referencing [ttrss-mobile#56](https://github.com/mboinet/ttrss-mobile/issues/56) and [ttrss-mobile#61](https://github.com/mboinet/ttrss-mobile/issues/61), which reference older devices.

The author did state that [he is no longer maintaining this project](https://github.com/mboinet/ttrss-mobile/issues/69), but if it does still work on normal/up-to-date devices then I'll add it back.

nodiscc:
@Kickball I have push access to mboinet's repo (at the time I did some cleanup in the way files/releases were organized). I don't have enough time or skills to work on feature that I have no personal interest in, but I think mboinet and myself would gladly accept patches.

About older devices, same problem, I don't own them, have no way to test, lack skills and time. I think the software can live on with those "bugs" (which could as well be platform bugs). I'd appreciate if you can leave ttrss-mobile in the list.

Otherwise you patch looks fine

Kickball:
I've readded ttrss-mobile in 5a796a7, given you have push access and can ensure it is maintained. :+1:

nodiscc:
https://github.com/mboinet/ttrss-mobile/issues/70 might take some time (we need an agreement from all contributors)

grittt-rss in fact uses the [BSD license](https://opensource.org/licenses/BSD-2-Clause)

AndyR207:
@Kickball I've resolved the merge conflicts here and we don't seem to be waiting on anything else...

nodiscc:
:+1:

PR #418, opened 2016-01-25
add Klaus Git repository viewer and ISC license

nodiscc:
ISC is functionally equivalent to the 2-clause BSD license: https://en.wikipedia.org/wiki/ISC_license


AndyR207:
Looks good from here.

PR #421, opened 2016-01-26
Remove dovecot from webmail client list

mburns:
It is already (correctly) listed under `Mail Delivery Agents`


AndyR207:
Good catch, thanks.

PR #439, opened 2016-02-08
Update contributing guidelines.

AndyR207:
Update contributing guidelines.


nodiscc:
What is the difference between Demo and Example?

AndyR207:
> By the way I have no strong feelings about proper capitalization.

The regex has fairly strong feelings about it though.

Kickball:
7b43cce removes the only use example use case, it can now be completely removed from the guidelines.

We should stick to strict capitalization as per the regex and guidelines.

AndyR207:
893ce1c removes "Example" from the guidelines.

nodiscc:
Looks good, thanks

PR #449, opened 2016-02-16
updated agendav with source, lang and license

cavebeat:


README.md
```diff
@@ -140,7 +140,7 @@ See also [Static Site Generators](#static-site-generators), [Content Management
 
 Some [Groupware](#groupware) solutions also feature calendar/address book editing and synchronization.
 
-  * [AgenDAV](http://agendav.org/) - Multilanguage CalDAV web client with a rich AJAX interface and shared calendars support
+  * [AgenDAV](http://agendav.org/) - Multilanguage CalDAV web client with a rich AJAX interface and shared calendars support [Source Code](https://github.com/adobo/agendav) `GPLv3` `JavaScript` `PHP`
```
nodiscc: PHP only. JS is only used client side in the browser. PHP is the required platform on the server.

cavebeat:
It's not stated in the Contributing section to list *only* the required platform. 
> main programming language/platform referenced.

Furthermore it's interresting to know the main language of the program and the required platform.

nodiscc:
Hi, it is expected (I think) for a web application to use HTML/CSS/JS.

What's interesting is to reference in this list are the approximate server requirements (eg. Node.js often equals running full Node stack for an application, PHP or Python web apps most likely require a webserver + PHP/WSGI stack, Go is likely a static binary, etc.) This is not true at all times of course (eg Reddit requires PgSql, Cassandra, RabbitMQ, redis/memcached....). But all of them use HTML/CSS/JS.

Else many other entries need to be changed to `Javascript` `PHP/Py/C...`

Is this ok?

cavebeat:
hmm, k, makes sense, i'll update

cavebeat:
hmm, otherwise, sometimes its nice to know if its html5 or JS or... it's just more info

nodiscc:
Thanks for the fix! The list is already dense so more info not always a benefit (_96.2 KB text file_); you can always follow the link to the project page to know what framework they used to develop the frontend, what clients exist, etc.

I agree that some entries have useless language precisions (Python/Django) @Kickball ok change these to just `Python`?

Kickball:
@nodiscc @AndyR207 Yep, we're alright to remove the useless precision which nodiscc mentioned above apart from with `Node.js` vs `Javascript`.

PR #450, opened 2016-02-16
updated baikal with source, lang and license

cavebeat:


nodiscc:
Thank you

AndyR207:
Please add a full stop after the description.

nodiscc:
@AndyR207 @Kickball I don't understand why this can't be merged? What's the exact problem with the full stop (you were talking about a regex, which one?) We can add full stops in batches afterhand. The missing full stop has been slowing down every patch for some time.

PR #452, opened 2016-02-18
fix tiddlywiki language, source, license

nodiscc:
(tw can work as a single HTML/JS file in a web browser, or as a Nodejs application)


Kickball:
Can we change the `Javascript/Node.js` to just `Node.js` or does it use pure JS as well?

nodiscc:
You can access (open tiddlywiki.html, single html+css+js file) it in a simple web browser without any Nodejs setup/install/requirement. So I think both languages qualify. Or change `Javascript` to `HTML5`. I've never used the Nodejs install mode myself.

Kickball:
We can stick with the current formatting.

PR #455, opened 2016-02-18
added Thingspeak, IoT Data Storage

cavebeat:


AndyR207:
Looks good to me.

nodiscc:
Needs to be moved to the Automation section, right?

> Send sensor data to the cloud.
> Analyze and visualize your data.
> Trigger a reaction.

AndyR207:
You also have an extra bracket before Source Code in this PR.

cavebeat:
i have put it into this section, because it fits to Freeboard. 
Freeboard is a visualizer for the data storage. I don't see myself something related to automation. 

so where to put it?
furthermore is phant no automation.

nodiscc:
Well from the description it looked more like Huginn (in Automation) but Personal Dashboards is ok.

Kickball:
I agree with @nodiscc that it does seem to be both Automation and Personal Dashboards, but until we use a tagging system we'll just add it to the Personal Dashboard section.

PR #458, opened 2016-02-18
update mailcow

cavebeat:


AndyR207:
Can only have one language. Please remove Shell.

nodiscc:
@AndyR207 Nope it can have 2, but separated by /
@cavebeat I fixed it for you at https://github.com/cavebeat/awesome-selfhosted/pull/1, please merge then I think we can proceed

PR #462, opened 2016-02-20
Add eBooks Category and Calibre (eBooks)

Kickball:


nodiscc:
I think calibre is a desktop application. If this is the case it can't be added. @Kickball can you confirm?

AndyR207:
I agree with nodiscc, Calibre looks desktop based.

AndyR207:
Closing this because it doesn't really fit into the list.

PR #466, opened 2016-02-22
Added Tuber - Peer-to-peer video chat that works

smacz42:
https://blog.trailofbits.com/2015/12/15/self-hosted-video-chat-with-tuber/

https://github.com/trailofbits/tubertc/blob/master/LICENSE


AndyR207:
Please move this before "Zulip" so it is correctly ordered.

AndyR207:
There needs to be a space between the name and the description.

AndyR207:
Looks good to me.

PR #478, opened 2016-03-03
added ShareLaTex

cavebeat:


AndyR207:
Looks alright to me.

Kickball:
There are brackets missing around the source code tag. See Turtl for an example of what I mean.

Kickball:
Looks good to me now!

PR #479, opened 2016-03-06
Add Kirby to non-free.md

fnkr:
The CMS I use on my website. I like it very much.


Kickball:
It doesn't fit perfectly with our standards, but then again nothing does in the non-free.md. We will have to fix the whole file at some point.

fnkr:
@Kickball I have no problems paying for software, as long as the source code is open.

Kickball:
When I said it doesn't fit perfectly to our standards I was refering to the formatting standards, which vary wildly between the readme.md and non-free.md lists. My apologies for the confusion.

PR #485, opened 2016-03-07
Added Kinto storage service

leplatrem:


nodiscc:
Thanks @leplatrem It looks like Javascript is only used client-side. If this is the case please change the language to just `Python`

leplatrem:
Yes indeed, I'll change it then.

leplatrem:
Updated, thanks!

AndyR207:
Please sort this entry into alphabetical order.

leplatrem:
Done!

Kickball:
Merged, thanks for the commit!

PR #490, opened 2016-03-09
Add eLabFTW to the list

NicolasCARPi:
Hello,

This PR is to add elabftw to the list.

Regards,
~Nico


nodiscc:
Thanks. I don't know if CMS is the best category though.

NicolasCARPi:
Yeah I'm not sure either. But as you don't have an ELN category, I put it here. Feel free to move it :)

nodiscc:
On the second look it indeed _is_ a CMS, with a specific purpose. The [ELN](https://en.wikipedia.org/wiki/Electronic_lab_notebook) article on Wikipedia for those curious about it. Good to merge, merci!

NicolasCARPi:
De rien ;)

Kickball:
We don't list the database backend used, so please remove the /MySQL from the language section.

nodiscc:
True. Thanks for the review @Kickball

NicolasCARPi:
Removed.

nodiscc:
Please use AGPLv3 like every other AGPL entry in the list. Squashing your commits would be appreciated.

NicolasCARPi:
AGPLv3 added, commits squashed :)

PR #492, opened 2016-03-09
Added "Manager" to Money+ section

MichaelTunnell:
Manager is a self-hosted accounting app using Mono and costs $199.

http://www.manager.io/server


nodiscc:
Link should be http://www.manager.io/server

MichaelTunnell:
agreed, I've updated the link as suggested.

nodiscc:
Thank you.

MichaelTunnell:
Nice! Glad I could help with the list, this is a great resource.

PR #495, opened 2016-03-10
Added Email Marketing category and phpList app

samtuke:
Disclaimer: I hack on and work for phpList ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,6 +231,12 @@ _Mailing lists servers._
   * [Mailman](https://www.gnu.org/software/mailman/) - The Gnu mailing list server. `Python` `GPL`
   * [Sympa](https://www.sympa.org/) - A mailing list manager. `Perl` `GPL`
 
+#### Email marketing
+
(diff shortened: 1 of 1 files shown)


samtuke:
@nodiscc phpList is not a mailing list manager in the usual sense of the term - mailing lists are for many-to-many communication, whereas phpList, and the hundreds of apps like it, are for one-to-many communication. Listing phpList as a mailing list application would, I fear, encourage confusion.

nodiscc:
@samtuke I understand. I think we should try to avoid creating new sections that contain only one entry (the table of contents is already large), but in this case _Mailing lists_ would be confusing.

Is it ok to change the section title to _Newsletter management_? You should also add a section link in the TOC.

samtuke:
@nodiscc Sounds good -- changed the section name to _Newsletter Management_, and added link in TOC.

nodiscc:
Thanks!

PR #497, opened 2016-03-10
Fix the language of bugzilla

tillarnold:
bugzilla is not written in [pearl](https://en.wikipedia.org/wiki/PEARL_(programming_language)) but in perl.


tillarnold:
Sorry I messed up with the branches. Will resubmit.

PR #502, opened 2016-03-12
Add Headphones - a music downloader

vrutkovs:
Not sure about the exact category for this - its not a player 
and barely a file sync software, please advice


nodiscc:
There is another downloader in Automation (Alltube). I think you should put Headphones there.

vrutkovs:
Right, Automation seem to fit better. Updated the PR

PR #509, opened 2016-03-16
Add Dashing to Personal Dashboard list

mide:
I recently came across [Dashing](https://github.com/Shopify/dashing), which is a fantastic dashboard-building tool.

### DIFF
--- a/README.md
+++ b/README.md
@@ -658,6 +658,7 @@ See https://github.com/n1trux/awesome-sysadmin#monitoring and https://github.com
 
 ## Personal Dashboards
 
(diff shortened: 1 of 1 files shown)


AndyR207:
Missing a full stop at the end of your description, other than that you're good.

mide:
Since this is a list and the descriptions aren't sentences, the need for the full stop could be argued. 

But that's a whole different issue, so I've gone ahead and added the period.

Thanks!

AndyR207:
The full stops are for compliance with our scanner that uses regex and expects them (and will thus throw up errors if they are missing).

Looks good to me.

nodiscc:
@AndyR207 the missing full stop should 1. either be tolerated by your scanner 2. either be added by yourself or one of the maintainers after the fact (you can edit the README). It should not require the submitter to edit his PR since the entry is valid (Since this is a list and the descriptions aren't sentences, descriptions missing it are valid)

PR #516, opened 2016-03-28
Adding Metabase

salsakran:
Created a new category under Analytics as we strictly speaking a web analytics tool.


salsakran:
Added the source code link.

Not sure really belongs in "business intelligence". I struggled a bit with the section. Think they fall more into "personal" or "social" analytics whereas we're used in a work context.

PR #519, opened 2016-03-30
Add Hospital Run

Kickball:
Two main discussion points for this PR: category and source code formatting.

### Category 

I felt this best fit into the Money, Budgeting and **Management** category, however there may be more fitting categories I missed. Also as there is no other similar software, I felt it doesn't warrant its own category.

### Source Code Formatting

This projects source code is split into two categories (front end and back end), this is not something we have encountered before and therefor requires discussion. The current formatting is only provisional, until we can agree on a format for multi repository projects.

What are your opinions on multi repository formatting @nodiscc @AndyR207?


nodiscc:
I'd just reference the backend source code since it has a clear link to the frontend dependency in it's README. Links to source code should try to point users to setup instructions, and provide a starting point for research/evaluation of the codebase.

> this is not something we have encountered before and therefor requires discussion.

Actually there are already a few projects that use multiple repositories in the list (can't remember exactly which ones right now). We just linked to the "main"/server-side repository and expect users to do some research if needed.

Using the management category is fine by me.

AndyR207:
I agree stick to the backend only.

Kickball:
The source code has been amended as per our decision above. 

I understand that we have already added projects which have separate repositories, however no discussions have happened in regards to this so I wanted to gather contributors opinion.

PR #520, opened 2016-03-31
Changed the pydio adress from pyd.io to pydio.com

runical:
Pydio moved from pyd.io to pydio.com and tries to redirect everyone who visits pyd.io. I updated the address because of that.


Kickball:
Thanks for this catch, this pr will be merged shortly.

PR #522, opened 2016-04-01
Add Hexchat to list

archaeme:


nodiscc:
> GTK+ IRC Client

This is a desktop application and therefore doesn't fit into this list.

PR #542, opened 2016-04-21
Adds Linux Game Server Managers

r0cknfire:
The tools are free, but the games are not. Should it be added to `non-free.md` instead ?


nodiscc:
Hi,

> The Linux Game Server Managers are command line tools for quick, simple deployment and management of various dedicated game servers and voice comms servers.

These are command line administration utilities and, I think, do not fit the kind of software we list here

>  network services and web applications which can be hosted locally.

r0cknfire:
My bad. I shall close this then.

PR #544, opened 2016-04-24
Add Puffin

loomchild:


loomchild:
In addition to adding my project, I fixed the Makefile. Please let me know if you'd like to split the pull request into two separate ones.

Kickball:
Please order Puffin alphabetically within the category. 

I don't see any problems with the MakeFile, but @nodiscc originally set it up.

nodiscc:
The makefile change is welcome. The recommended syntax in CONTRIBUTING specifies `[Demo](...)` **before** `[Source Code](..)`. Thanks for fixing this.

After fixing the alphabetical ordering it should be good to merge.

@loomchild If you are a user of Puffin and want to share two words about how it works (it's not very clear to me, automated deployment of services on a Linux machine? Somehow like Sovereign?) or what you like about it, feel free to share.

Edit: I get it https://github.com/puffinrocks/puffin/blob/master/CONTRIBUTING.md#running-your-own-puffin. It's distributed as a Docker image. The website at http://puffin.rocks/ is for their _hosted_ solution.

loomchild:
Hi @Kickball and @nodiscc,

First of all sorry for late reply.

In fact the Makefile fix is mostly about brackets - when I created the entry I noticed that square brackets were generated, instead of normal brackets. Then I realized that Demo and Source Code links are inverted as well.

About Puffin - I am actually develop the project. The idea is to help people self-host their software. The main difference between other similar private-cloud projects such as Cozy.io or Sovereign, is that it uses containers, which isolate applications from each other (imperfect yet, but we are getting there), are technology / programming language agnostic, and permit to cheaply host other people's applications (not 100% self-hosted ideology I realize, but my thinking is that it's nothing wrong in hosting software for friends, as it could be a first step in moving away from centralized services, such as Facebook, Github, etc.) Another benefit of Docker is that a lot of existing software is already containerized, so adding new applications to the catalog is relatively easy.

The http://puffin.rocks website is more of a demo right now than a hosted solution, but contains some user friendly explanations and links to code, blog and forum and is intended as a central source of information.The self-hosted part is explained in the [About section](http://puffin.rocks/about.html). If you prefer to link to Github page as a main project page, I can do it too. 

I corrected the ordering, I though only the authors are alphabetical and the projects are FIFO.

PR #547, opened 2016-05-03
fix a typo

marclaporte:


nodiscc:
English is not my native language but I'm pretty sure `it's` is correct. @Kickball @AndyR207 ?

marclaporte:
Not in this case.
Please see: http://www.merriam-webster.com/words-at-play/when-to-use-it's-vs-its

Thanks!

nodiscc:
> use it's only when you mean it is or it has

ok.

PR #567, opened 2016-06-15
add unison to file synchronization

ryantm:
Unison is a very nice file synchronization program licensed under GPL.


Chouchen:
Isn't it more a desktop app than an hosted one?

nodiscc:
Unison has both a client and a server-side part (even though it uses SSH to communicate between the client and the server).

PR #570, opened 2016-06-17
Add Gigablast

rangeroob:
Add gigablast open source search engine


nodiscc:
Thank you, ready to merge :+1: 

@rangeroob can say a few words about your experience with this, what you like/dislike in this project, etc? (how does it compare to other FOSS search engines, etc)

rangeroob:
The only other self hosted search engine i have had experience with is yacy so I'll try and compare the two. One Gigablast is faster hands down love it for the speed that it provides over yacy. Two going off of both public instances Gigablast seems to have more results and gives you the option to go to google, bing and even amazon. In my mind it feels similar to duckduckgo. The one down side i noticed is that the requirements especially 4gbs of ram unlike yacy which can be run on much lower.

PR #571, opened 2016-06-17
Sphinx

rangeroob:


nodiscc:
@rangeroob This PR also adds Gigablast, which you also add in https://github.com/Kickball/awesome-selfhosted/pull/570. Please work on a separate branch. Also I'm not sure whether we should include Sphinx since it's meant to be integrated in other applications, and is not directly usable by itself.

rangeroob:
@nodiscc sorry about the mess up and i agree your decision not to include sphinx in the list.

PR #584, opened 2016-06-24
Replace unmaintained Shout with its official community fork

astorije:
The Lounge is a community initiative started earlier this year, after development on Shout stopped.
Demo looks sloppy at the moment, I know...

See https://github.com/erming/shout#official-forks


nodiscc:
Thank you

PR #69, opened 2015-07-30
Added static site generator / blogging engine Hugo

trenton42:

PR #587, opened 2016-06-24
Add Hugo to `README.md`

dslachut:
Hugo is a fast, easy to use, and actively maintained Static Site Generator that I use for hosting various sites.


nodiscc:
Hi, we don't list static site generators. See https://github.com/Kickball/awesome-selfhosted/pull/69 -> https://github.com/Kickball/awesome-selfhosted#static-site-generators. Thanks for the effort though.

Issue #471, opened 2016-02-26
Missing Licenses


Kickball:
The following issues where identified in awesome-selfhosted by @AndyR207:

    [Countly](https://count.ly) - Real time mobile & web analytics, crash reporting and push notifications platform ([Source Code](https://github.com/countly)) `AGPL` `Javascript`
- [x]  Demo needs to come Before Source Code

    [Bookie](https://github.com/bookieio/Bookie) - Python based delicious.com replacement. ([Firefox addon](https://github.com/bookieio/bookie-firefox), [Demo](http://bmark.us/), [Documentation](http://docs.bmark.us/) `unlicensed` `Python`
Firefox Addon and Documentation are not valid. Please remove them. Please capitalize "Unlicensed" although I feel like this entire entry should be removed because I don't think we're accepting "Unlicensed" as a valid thing?? @Kickball @nodiscc AGPLv3, fixed -- nodiscc

    [Firefox Sync Server](https://docs.services.mozilla.com/howtos/run-sync-1.5.html) - Sync Firefox bookmarks, passwords, history, tabs, preferences. ([Source Code](https://github.com/mozilla-services/syncserver)) `unlicensed` `Python`
See above comment about licensing. -- MPLv2, fixed -- nodiscc

    [Lobsters](https://lobste.rs) - Run your own link aggregation site. ([Source Code](https://github.com/jcs/lobsters))
- [x] No license 
- [x] No language

    [Shaarli](https://github.com/shaarli/Shaarli) - Personal, minimalist, super-fast, no-database bookmarking and link sharing plaform. ([Demo](http://shaarlidemo.tuxfamily.org/Shaarli/) `demo/demo`) `Zlib` `PHP`
- [x] Remove the demo/demo part -- these are the demo credentials and they should be kept -- _nodiscc 2016-02-27_

    [calypso](http://keithp.com/calypso/) - A python-based CalDAV/CardDAV server. ([Source Code](http://keithp.com/cgi-bin/gitweb.cgi?p=calypso.git;a=tree)) `GPLv3`
- [x] Missing language

    [DAViCal](http://www.davical.org/) - A server for calendar sharing (CalDAV). ([Source Code](https://gitlab.com/davical-project/davical)) `MIT`
- [x] Missing language

[...]

PR #591, opened 2016-06-25
Update Exim Missing Information

Kickball:
* Add Exim Language
* Add Exim Source Code
* Update Exim Site Link to use HTTPS

Resolves part of #471.


nodiscc:
Same as #592, I'd rather link to the official source repository at http://git.exim.org/exim.git. The github mirror can be added as a secondary source code link like we did with ArkOS.

Kickball:
@nodiscc I've updated the license and source code link as suggested above.

nodiscc:
Everything is ok.

PR #592, opened 2016-06-25
Update OpenSMTPD with missing information

Kickball:
* Add OpenSMTPD Language
* Add OpenSMTPD Source Code

Resolves part of #471.

Note: this is alternate link for the source code: http://cvsweb.openbsd.org/cgi-bin/cvsweb/src/usr.sbin/smtpd/.


nodiscc:
I'd rather link to the official CVSweb repository, the github repository is just a mirror. @Kickball If you want you could put both source links in the entry, like we did with ArkOS (https://github.com/Kickball/awesome-selfhosted#self-hosting-solutions)

Kickball:
I've updated the source code link with the official CVSweb repository.

PR #598, opened 2016-07-01
Add Keeweb

cristianmenghi:
A web based password manager compatible with keepass


nodiscc:
Ah, I see keeweb has changed URLs. Can you update the existing entry instead? (top of that very same section ) Thanks.

nodiscc:
I think you closed the PR by mistake. It was all good.

cristianmenghi:
Hi, im fixit the url of keeweb and remove the duplicate entry i made.

PR #484, opened 2016-03-07
Add MkDocs

rrerolle:

PR #601, opened 2016-07-02
Added MkDocs

Aisbergg:
A static site generator for building project documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,7 +291,7 @@ _Mass mailing software - one message to many recipients_
 ### IPBX
 
  *[IPBX](https://en.wikipedia.org/wiki/IP_PBX) telephony software*
- 
+
   * [Elastix](http://www.elastix.org) - Unified communications server software based on Asterisk. ([Source Code](https://svn.code.sf.net/p/elastix/code/trunk/)) `GPLv2` `PHP`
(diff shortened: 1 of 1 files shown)


nodiscc:
> A static site generator

We don't list them: https://github.com/Kickball/awesome-selfhosted#static-site-generators

https://github.com/Kickball/awesome-selfhosted/pulls?q=is%3Aclosed%20static
https://github.com/Kickball/awesome-selfhosted/pull/484

You can use the search functionality to check if an entry has already been submitted.

Aisbergg:
In the category `Documentation Generators` are more of those static site generators listed. *MkDOcs* follows the same principals as *ReadTheDocs* and *MdWiki*. It also comes with an integrated server, that is looking for file changes and serves the generated conetet.

nodiscc:
@Aisbergg the goal is to avoid duplication of effort. As you can see MkDocs and Mdwiki are already listed at https://staticsitegenerators.net/ so it's ok to remove those from here (can you do it?). I'd agree to also remove ReadTheDocs, but it needs to be ported to staticsitegenerators.net first (https://github.com/bevry/staticsitegenerators-list/edit/master/list.yaml)

Aisbergg:
@nodiscc I will take care of it. I think a reference to the `Static site generators`  in the `Documentation Generators` section would be reasonable, because most user expecting the documentation apps to be there. (At least I did)

Aisbergg:
See #620

PR #604, opened 2016-07-03
Adding WikiSuite

marclaporte:


jgillich:
@marclaporte I checked out WikiSuite and I'm confused..what exactly is it? There is no code, so it's not software. Is it simply a project to make hosting stuff on ClearOS easier by providing documentation and contributing to the individual projects?

marclaporte:
@jgillich  Good question! So far, we've been able to add / improve code directly on upstream projects, and avoid forking, hacks or plugins, and their inherent problems: http://pluginproblems.com/

The contributed code is in these 3 categories:
1- Add / improve features (useful even if you only use one component, and have no immediate need for WikiSuite)
2- Add / improve interoperability with the other components, to make the whole thing act as one app (good integration). For example, we've been working with various projects to move them to Bootstrap, to make it easier to have a consistent UX: http://wikisuite.org/Bootstrap
3- Make the components easier to install and upgrade (mostly via RPMs which can be used on CentOS as well), and next we are going to create ClearOS apps, with GUIs.

Best regards,

M ;-)

PR #617, opened 2016-07-08
Add ClickHouse

pelim:


nodiscc:
It looks like a command-line data processing tool, technically this is not a network service and therefore doesn't belong to this list. https://clickhouse.yandex/tutorial.html. Interesting nevertheless. I think there is an awesome-bigdata list out there, this might interest them.

PR #620, opened 2016-07-10
Removed 'daux.io' and 'MdWiki' from 'Documentation Generators'

Aisbergg:
Removed *daux.io* and *MdWiki* from `Documentation Generators´, because they are in fact static site generators. Kept *Read the Docs* in the list, because it is not really meant to generate static site content even though it relys on *Sphinx* (little inpractical this way). It is more about collecting docs from several repositories and generate different content (HTML, PDF, etc.) in a automized fashion.


nodiscc:
Thank you

PR #621, opened 2016-07-11
Add CraftCMS to the list of Content Management Systems

volodymyr-kushnir:
Add CraftCMS to the list of Content Management Systems


nodiscc:
Thank you, unfortunately it seems CraftCMS is non-free (proprietary) software, as per https://github.com/Kickball/awesome-selfhosted#contributing this entry must be added to https://github.com/Kickball/awesome-selfhosted/blob/master/non-free.md instead.

volodymyr-kushnir:
I'm not sure, I was able to install and use it, it's kind of freemium (free to use with single admin account). I'll double-check it later.

Kickball:
@volodymyr-kushnir did you check?

Kickball:
@volodymyr-kushnir we will close this PR 7 days from now due inactivity, and the limit of one admin account.

PR #627, opened 2016-07-18
Include netdata into Analytics

rolurq:
You should really consider this service. It's incredibly fast and easy to install.


nodiscc:
Hi, netdata is a server monitoring tool (https://github.com/Kickball/awesome-selfhosted#monitoring). We delegated system administration tools to https://github.com/n1trux/awesome-sysadmin.

Netdata can already be found here https://github.com/n1trux/awesome-sysadmin#metric--metric-collection, so there's nothing to add.

Thanks and feel free to send other patches.

PR #629, opened 2016-07-22
update demo link for Kopano

fbartels:
Hi, the demo link for Kopano was missing the http and is therefore broken. Fix attached.


nodiscc:
Thanks

PR #637, opened 2016-08-03
Added PHP QR Code

Ober7:


AndyR207:
Looks good to me.

nodiscc:
I think this should be removed. It's a library, not a web application. ping @Kickball @AndyR207

Kickball:
Yeah I think @nodiscc is right. We should remove this as it is not a web application.

nodiscc:
Done https://github.com/Kickball/awesome-selfhosted/commit/80aa2dadd681d1a0e59f3526965606f96238c51d

Issue #635, opened 2016-07-31
Add Webtrees


n8225:
Homepage: [https://www.webtrees.net/index.php/en/](https://www.webtrees.net/index.php/en/)
Source Code: [https://github.com/fisharebest/webtrees](https://github.com/fisharebest/webtrees)
webtrees is the web's leading on-line collaborative genealogy application.
GPLv3
PHP

I wasn't sure about creating a category for 1 item. Or possible a subcat?
edit: fixed links

nodiscc:
@n8225 your links' markdown syntax is broken.

It doesn't really fit in any other category. I think it's fine to Create a Genealogy category.

PR #638, opened 2016-08-04
Added webtrees issue #635

n8225:


README.md
```diff
@@ -46,6 +46,7 @@ See [Contributing](#contributing).
       - [Web based file managers](#web-based-file-managers)
   - [Games](#games)
   - [Gateways](#gateways)
+  - [Geneaology](#geneaology)
```
Chouchen: should be `genealogy` if we're creating a genealogy category

n8225:
Thanks for catching that.

Kickball:
@nodiscc @AndyR207 @n8225 

I don't feel that genealogy warrants its own category, as there are not many selfhosted genealogy solutions and I don't imagine that genealogy is very popular. However it doesn't fit into any of the existing categories.

We could look at creating a misc/other category for solutions that don't fit into any categories and are not popular enough to warrant their own categories. How do you all feel about that?

n8225:
I'm fine with either route.  I have found another project to add: [https://github.com/geneweb/geneweb](https://github.com/geneweb/geneweb)

n8225:
Now that I've thought about it a little more,  I'm leaning towards the misc/other category.   If we go that route we should also set a minimum number of entries for a new category.

nodiscc:
+1 please move this to a `Misc/Other` category. 3 items seem like a decent number to warrant creating a new section. You can also add this rule to the Contributing section.

n8225:
Tried rebasing and it looks like I may have made a mess.  If need be I'll just delete this branch and copy paste the changes to a new branch.

nodiscc:
Yes it  would be better to start a new PR. You can close this one when done.

PR #639, opened 2016-08-04
Agendav Fix

n8225:
Added missing parenthesis around Source Code for Agendav


Kickball:
Very nice catch, thanks for finding this!

PR #643, opened 2016-08-04
Fix Kanboard license

mlbarrow:
Kanboard is MIT-licensed, not AGPL3. Ref: https://github.com/kanboard/kanboard/blob/master/LICENSE


Chouchen:
Looks good to me. They switched to MIT in 2015.

Kickball:
Change looks good, nice catch!

PR #649, opened 2016-08-12
Added SupportPal

bytestream:
Paid alternative to osTicket


Kickball:
@nodiscc just a quick question before merging this, should this not be in awesome-sysadmin?

nodiscc:
@Kickball I don't think so, awesome-sysadmin moved their ticketing software here +      it's not administration software. + I don't really care about non-free.md :)

nodiscc:
Thanks

PR #655, opened 2016-08-18
Corrected Spacing and removed HL above Self-hosting Solutions

AndyR207:
Made spacing between sections consistent and removed the horizontal line above Self-hosted solutions.


nodiscc:
Thank you

PR #659, opened 2016-08-21
Add Medusa

OmgImAlexis:
Medusa is a heavily modified fork of Sickrage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,7 @@ _Business Intelligence_
   * [Healthchecks](https://healthchecks.io/) - Django app which listens for pings and sends alerts when pings are late. ([Source Code](https://github.com/healthchecks/healthchecks)) `BSD` `Python`
   * [Huginn](https://github.com/cantino/huginn) - Allows you to build agents that monitor and act on your behalf - `MIT` `Ruby`
   * [Http2pic](https://http2pic.haschek.at/) - Website screenshots/renderer. It uses the wkhtmltox to render websites with various options. ([Source Code](https://github.com/chrisiaut/http2pic)) `Apache 2.0` `PHP/Javascript`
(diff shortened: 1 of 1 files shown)


Kickball:
Medusa needs to be ordered alphabetically, please fix this.

OmgImAlexis:
It might be an idea to add [Github PR template](https://github.com/blog/2111-issue-and-pull-request-templates) so people know this before opening the PR.

Fixed btw.

nodiscc:
Thank you, I thought we had this requirement written down in [Contribution](https://github.com/Kickball/awesome-selfhosted#contributing), but apparently no. This should be fixed soon.

Kickball:
Just to clarify this is a fork of Sickrage right?

nodiscc:
It seems to be. I would be wary of adding too many forks.

Kickball:
As long as there are a few major differences and changes then I am happy to list it as its own item. However if it is still similar then I think it should be listed as a sub-item.

nodiscc:
@pymedusa Can you list the major differences between SickRage and Medusa?

fernandog:
@nodiscc

First and most important: user can manual search. It shows all available releases with all needed info like quality, seeders, leechers, color font when find ignored or preferred words.

Its the only sickbeard fork that does that.
Entire team of devs - except one - left SickRage.
So here we are.

fernandog:
Check it out:

image

Kickball:
Forgive me as I am not familiar with Sickrage and its interface, so I can't compare your screenshot. Can you clarify if the manual search is the major only change?

Kickball:
74765c1 adds Medusa, as agree with the collaborators in the Awesome Selfhosted chat.

PR #661, opened 2016-08-21
update link to invoiceninja for self hosted version

deserted:
current link is to the commercial hosted version, update to the open source self hosting option


Kickball:
Looks good, thanks for catching this.

Issue #660, opened 2016-08-21
Add Icecast 2 to Video streaming category as well


ePirat:
Please add the Icecast 2 streaming server to the video streaming category as well,
as it is capable of video streaming, not only audio streaming.

nodiscc:
@ePirat it is not advertised as suche

http://www.icecast.org/:

> Icecast is a streaming media server which currently supports Ogg (Vorbis and Theora), Opus, WebM and MP3 audio streams.
> It can be used to create an Internet radio station or a privately running jukebox and many things in between. It is very versatile in that new formats can be added relatively easily and supports open standards for communication and interaction.

ePirat:
The “audio streams” refers to MP3, not the other formats, after all Theora is a video codec. But I agree, that is phrased in a very confusing manner and I will make sure to change it.
I can confirm that Icecast is definitely able to stream video.

nodiscc:
Ok good to know. Can you change the description to `streaming audio/video server ` and duplicate the entry into the Video Streaming section? Thanks

Kickball:
Closed in #663.

PR #663, opened 2016-08-21
Add Icecast to video streaming section

ePirat:
This adds Icecast to the video streaming section as well, and clarifies that it is an audio and video streaming server in the description. (Ref. #660)


nodiscc:
Thanks

PR #664, opened 2016-08-21
Add FreeIPA to "Federated Identity/Authentication"

ilsi:


Kickball:
Would it not be better to change the license to various as we have done before?

nodiscc:
@Kickball I changed the wording in #669 to `Multiple` when there are multiple licenses.

Kickball:
Alright, should we change the licenses to `Multiple` then?

nodiscc:
Yes please @ilsi remove the link to the licenses page and simply tag it `Multiple`.

ilsi:
Ok. Retaged to `Multiple`

PR #682, opened 2016-08-22
Add Liferea feed reader

louisbourque:


nodiscc:
Liferea is very good, but a desktop application (not a web application or network service). Closing

PR #692, opened 2016-08-23
Add Para

albogdano:
I've added a new category because it didn't fit in any of the other ones.


Kickball:
Please move to misc/other category as per contributing guidelines.

> Minimum of 3 entries to start a new category. Otherwise add to the Misc/Other category.

PR #706, opened 2016-08-23
Add language tags on Feed Readers

remram44:


Kickball:
Looks good to me, thanks for this!

PR #726, opened 2016-08-27
Add Awesome Selfhosted Chat to external links

Kickball:


n8225:
Should the External links section be alphabetized?

Kickball:
I think its one of the (only) categories we would best benefit from prioritizing ourselves.

n8225:
Agreed

PR #736, opened 2016-08-30
Fix file sharing

n8225:
Issues #35 #471

Updated Licenses, languages, and links.

Thank you for taking the time to work on a PR for Awesome-selfhosted!

To ensure your PR is dealt with swiftly please check the following:

- [x] Your submissions are formatted according to the guidelines. 
        
    ``[Name](http://homepage/) - Short description. ([Demo](http://url.to/demo), [Source Code](http://url.of/source/code)) `License` `Language```
- [x] Your additions are ordered alphabetically.
- [x] Your additions are [free software](https://en.wikipedia.org/wiki/Free_software), or if not they have been added to [non-free](non-free.md).
- [x] Any licenses you have added are in our [list of licenses](https://github.com/Kickball/awesome-selfhosted/blob/master/README.md#list-of-licenses).
- [x] You have searched the repository for any relevant issues or PRs.
- [x] Any category you are creating has the minimum requirement of 3 items.

PR #738, opened 2016-08-30
Fix spacing issues identified in #736.

Kickball:


n8225:
Looks good, not sure how I missed those.

PR #748, opened 2016-09-02
Fix media streaming

n8225:
Issues #35 #471

Thank you for taking the time to work on a PR for Awesome-selfhosted!

To ensure your PR is dealt with swiftly please check the following:

- [x] Your submissions are formatted according to the guidelines. 
        
    ``[Name](http://homepage/) - Short description. ([Demo](http://url.to/demo), [Source Code](http://url.of/source/code)) `License` `Language```
- [x] Your additions are ordered alphabetically.
- [x] Your additions are [free software](https://en.wikipedia.org/wiki/Free_software), or if not they have been added to [non-free](non-free.md).
- [x] Any licenses you have added are in our [list of licenses](https://github.com/Kickball/awesome-selfhosted/blob/master/README.md#list-of-licenses).
- [x] You have searched the repository for any relevant issues or PRs.
- [x] Any category you are creating has the minimum requirement of 3 items.

PR #749, opened 2016-09-02
makefile: syntaxerrors: detect )`

nodiscc:
Detect weird character pairing found in https://github.com/Kickball/awesome-selfhosted/pull/748


Kickball:
I'm not too familiar with makefiles, but I'm going to guess this change works.

Issue #875, opened 2016-11-30
Validate Pull Requests with a Continuous Integration system


m1guelpf:
@Kickball Is there a reason for this repository to avoid using http://danger.systems ?

nodiscc:
Link is broken. What do you mean?

nodiscc:
Got it. http://danger.systems/ is the correct link. The question about a Continuous Integration system/tests has bee brought up before. See https://github.com/Kickball/awesome-selfhosted/issues/63

@AndyR207 has built a custom scanner which detects invalid entries. I think it's still work in progress, and still not public. I think it would be useful to have a CI system that integrates with Github, so that contributors are immediately aware of problems in their Pull Requests.

There are a few checks available in the [Makefile](https://github.com/Kickball/awesome-selfhosted/blob/master/Makefile) (`make checks`), and I think a draft of Travis integration has been pushed to a random branch, but I can't find the discussion about it. If you'd like to improve the makefile and/or add Travis tests I'd be glad to merge your changes.

m1guelpf:
To start, @BitQuote made Awesomebot to check and report broken links directly to pull requests.
Also, you can build custom checks in Ruby and integrate them with Danger to report directly to pull requests.

nodiscc:
The required features for such a CI system would be:

 - [ ] Detect broken links
 - [ ] Detect outdated projects
 - [ ] Detect syntax errors
 - [ ] Detect missing language/license
 - [ ] Integration with the Pull Request system (warnings)
 - [ ] Avoid extra complexity in the repository (a single .travis.yml + Makefile would be ideal)
 - [ ] Must run automatically
 - [ ] Rely on Free Open Source Software

So, mostly enforce contribution guidelines. I think @AndyR207's bot does most of these it just lacks Github PR integration, and a public source release + a running instance. Is there any ETA on this? Else anyone is welcome to propose an implementation (in a Pull Request).

n8225:
If you take a look at the https://github.com/Kickball/awesome-selfhosted/tree/addTravis branch, I have started working on it. I hope to put some time into it next week.

nodiscc:
Great. The list of relevant commits can be found here: https://github.com/Kickball/awesome-selfhosted/compare/addTravis

m1guelpf:
@nodiscc I will have a look. Anyway, Danger IS open source: https://github.com/danger/danger

nodiscc:
[...]

PR #892, opened 2016-12-10
Add travis

n8225:
Issue #875

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,8 +11,7 @@ before_install:
 before_script:
   
 script:
-  allow_failures:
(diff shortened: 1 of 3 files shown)


m1guelpf:
@n8225 Awesome!

nodiscc:
I don't have the required knowledge of JS/node to review the `test.js` script. Everything else looks good. 

@n8225 you may want to remove the legacy checks from the Makefile (and update documentation about it) since this makes them redundant.

I'm in favor or merging this, and see over time if it works properly.

Big thanks

n8225:
I've made a few changes I think this is good to merge now.  See #905  for an example PR.(AwesomeBot for some reason failed one link that works fine.  So I changed awesomebot to only warn on link issues.)

PR #1005, opened 2017-03-30
Remove PHP Password Manager and RatticDB

ardichoke:
Both project sites are down and neither has a commit in over 6 months.


nodiscc:
Thanks @ardichoke

n8225:
Thanks @ardichoke

PR #1133, opened 2017-08-03
Add Virtualmin to the list of complete email solutions

gjedeer:
Virtualmin served great before I learned how to configure Postfix and Dovecot myself. It's free (GPL), it has been there for years, and the support is very good.

http://webmin.com/virtualmin.html


nodiscc:
Already listed at https://github.com/n1trux/awesome-sysadmin#control-panels

Issue #1038, opened 2017-04-21
Use a machine-readable format to hold the list data


ghost:
An interesting idea would be to create an XML document containing all repository data. Then someone could create an XML viewer, and then there would be an easy way to offline view the raw data of the repo. Currently, the only way that this data could be viewed offline, is via downloading the README, and users would not be able to view it due to it being written in markdown. I could start creating the file. Thoughts?

Kickball:
We have had similar ideas, but often with YAML in mind. We did have a project underway that transformed the list into many different formats (https://github.com/arylatt/awesome-selfhosted-com).

cpdevelops:
Is that project still going, and if so would it be ok for me to start contributing to it?

Kickball:
I don't believe the project has been worked on for a while, and I don't think @arylatt or myself will be able to help with it. You are welcome to start contributing to it, at the moment the documentation is not super complete.

The most up to date branches are data-parsing and data-parsing-2, I'd start by looking at them.

cpdevelops:
Ok thanks for the info. I'll close the issue now and get to work on that.

Kickball:
@cpdevelops happy to hear that you're interested in moving it forward, the best way to discuss it with @arylatt and I is via our rocket chat (https://chat.awesome-selfhosted.com).

nodiscc:
https://github.com/kkismd/pandoc-ymltbl can generate HTML/Markdown tables from another format (YAML). I'll have a try soon with a basic subset of the list to see if it could work for us.

Another aspect is the viewer, it must support filtering/sorting the list/table. If we're talking about viewing the list in a web browser there is probably a javascript library that can do that.

Reopening and renaming issue for further discussion

nodiscc:
I have started working on something at https://github.com/nodiscc/awesome-selfhosted-yaml (short yaml template), it think the schema will be able to cover pretty much all our needs. I will try converting it to multiple display formats sometime soon. Pull Requests welcome.

nodiscc:
As discussed in https://github.com/Kickball/awesome-selfhosted/issues/1200, each list item may have `last_update` and `maintained` keys, to allow for easy filtering of old or unmaintained projects.

Unmaintained projects could be kept in the raw list, while being hidden by default in the render

nodiscc:
From #1287

- [ ] List non-web clients in the yaml file

```yaml
name: 'Mpd'
homepage_url: 'https://www.musicpd.org/'
[...]

PR #1213, opened 2017-10-14
Added aenigma

nikksno:
https://github.com/openspace42/aenigma

The license is very close to "the unlicense", it's 100% public domain, hope that works.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -406,7 +406,7 @@ _[Extensible Messaging and Presence Protocol](https://en.wikipedia.org/wiki/XMPP
 
 #### XMPP Servers
 
-  * [aenigma](https://github.com/openspace42/aenigma) - The | state-of-the-art | secure-by-default | one-touch-deployed | XMPP server for everyone.  `LibreLicense` `Erlang [ejabberd]`
+  * [aenigma](https://github.com/openspace42/aenigma) - The | state-of-the-art | secure-by-default | one-touch-deployed | XMPP server for everyone.  `Unlicense` `Erlang`
   * [ejabberd](https://www.ejabberd.im/) - XMPP instant messaging server. ([Source Code](https://github.com/processone/ejabberd)) `GPL` `Erlang`
   * [Kontalk](http://kontalk.org/) - Kontalk is an Open Source Messenger, similar to WhatsApp (app for android only currently), including end-to-end encryption, server is based on Tigase XMPP Server. ([Source Code](https://github.com/kontalk)) `GPLv3` `Java`
   * [Metronome IM](https://lightwitch.org/metronome) - Fork of Prosody IM. ([Source Code](http://code.lightwitch.org/)) `MIT` `Lua`


danger-awesome-Selfhosted:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Error">
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Found 1 entries with syntax error(s).</td>
    </tr>
  </tbody>
</table>

#### Syntax Issues

 Name | Entry
----|----------------------
aenigma |   * [aenigma](https://github.com/openspace42/aenigma) - The | state-of-the-art | secure-by-default | one-touch-deployed | XMPP server for everyone.  `LibreLicense` `Erlang [ejabberd]` 
 

<p align="right" data-meta="generated_by_danger">
  Generated by :no_entry_sign: <a href="http://danger.systems/">Danger</a>
</p>

nikksno:
I've looked into the travis log but can't understand what I've done wrong. Please help. Thanks! ;]

nikksno:
Hi there. Thanks for your reply. I understand your point. I just think that since a "pure" ejabberd installation [the entry right below the would-be aenigma entry] is entirely insicure by default, whereas aenigma has server hardening with debian-first-boot-script [ssh, fail2ban, ufw, etc...], state of the art config, letsencrypt certs auto install, nginx reverse proxy to the admin interface, and so on, the benefits outweigh the costs, in my opinion. In any event, if this is a guideline, I agree it should be followed, and not broken [again] for such a small project. Thanks for looking into this.

nodiscc:
> I just think that since a "pure" ejabberd installation is entirely insicure by default

Can you explain why? aenigma's README doesn't explain what problems it addresses. Also have you considered sending patches/reporting bugs upstream (ejabberd or ubuntu in that case)?

Closing for now; but feel free to provide more info, thanks.

nikksno:
Not insecure in the sense of bugs or poor design, but just as a standard install of postfix or dovecot doesn't nearly have any kind of security by default in mind, ejabberd as a clean install via apt or dpkg has no security impli

nodiscc:
> ejabberd as a clean install via apt or dpkg has no security by default in any way

Yet you didn't detail _why_ exactly (here or in the project README). Are there links to upstream bug reports?

> aenigma takes server hardening and state of the art configurations to heart

Yet it doesn't address basic hardening such as SYN cookie protection ([example sysctl config](https://github.com/nodiscc/srv01/blob/master/roles/common/files/etc_sysctl.d_ipv4-security.conf)) and many others :)

> mailinabox is a perfect example of the same kind of ideal

A full mail server setup is _extremely_ tedious and difficult to configure properly, I believe that's the main reason why Mail-in-a-box exists (from their homepage). Installing and configuring ejabberd is much easier.

> it's still a bash script
> eventually migrating it to an ansible recipe
> looking for others interested in making the project better

My suggestions:

* Look for contributors in communities such as https://www.reddit.com/r/selfhosted
* I like bash scripts for many uses but this isn't one of them. Shell scripts can fail or behave wrongly in many unexpected ways, there is no idempotence, etc (but yes `set -e` is a good start. What happens if the script fails in the middle of execution and I have to re-run it? Will it work? etc.). There are many subtleties which make it easy to mess things up. Look at the [shellcheck report for aenigma](https://gist.githubusercontent.com/nodiscc/19b7d98f3feb63d38166e3d1069f685a/raw/8136b0a62fbcb879354672e07bcb139d77c35737/aenigma-shellcheck). There are similar warnings for dbfs.
* Try to send a patch to one of the projects at https://github.com/Kickball/awesome-selfhosted#self-hosting-solutions to adapt them to your needs
* Move to ansible/config management: one of the advantage of CM playbooks is that it makes it easier to determine *what* your configuration process does ([example](https://github.com/nodiscc/srv01/blob/master/roles/prosody/tasks/prosody.yml) with prosody).
* Get rid of that `wget -qO dfbs https://gh.nk.ai/dfbs && sudo bash dfbs`. https://github.com/openspace42/Debian-First-Boot-Setup is really easy to implement directly in your tool (just a couple of package installation and sshd configuration actually...)
* Try to report and fix the default ejabberd insecure (?) configuration in Debian or Ubuntu, or upstream.

nikksno:
Thanks for your suggestions.

    Yet you didn't detail why exactly (here or in the project README). Are there links to upstream bug reports?

    mailinabox is a perfect example of the same kind of ideal

The 2k+ lines of code [which could probably be a lot less if the code was DRYer] are the evidence - in my opinion - that configuring a secure instance of ejabberd [like basically any other service on the planet] is extremely hard do to correctly and with all services playing well with eachother. The **features** section lists why aenigma is better than a pure ejabberd install.

    Yet it doesn't address basic hardening such as SYN cookie protection (example sysctl config) and many others :) [1] [2]

Great to know thanks, I'll I'll look into it ASAP.

    What happens if the script fails in the middle of execution and I have to re-run it? Will it work?

Absolutely. aenigma *is meant* to be re-run and warns in case of incomplete installs

    There are many subtleties which make it easy to mess things up. Look at the shellcheck report for aenigma. There are similar warnings for dbfs.

That's on my to do list, thank you.

    Get rid of that wget -qO dfbs https://gh.nk.ai/dfbs && sudo bash dfbs.

The installer actually does a git clone now directly inside aenigma

    Try to report and fix the default ejabberd insecure (?) configuration in Debian or Ubuntu, or upstream.

Again, ejabberd is perfectly secure in terms of code and goals, but just like it's not anyone's goal to make a postfix or dovecot secure by default [hence the existence of mailinabox], ejabberd as an apt install is definitely not secure or entirely usable by default for my strict standards. Every line of the aenigma installer is - for me - a *requirement* for an instance of an XMPP server that I would use.

nikksno:
Extra question: how do you quote and indent in markdown? ;]

nikksno:
https://github.com/openspace42/aenigma/issues/27

nodiscc:
My point is that a configuration script for a single service and a base OS setup is not enough to warrant an entry in https://github.com/Kickball/awesome-selfhosted#self-hosting-solutions (yes Mailinabox is borderline). Note that other projects listed there are "platforms" that ease up installing many services/applications.

Maybe there's room somewhere for such scripts/playbooks (https://github.com/Nyr/openvpn-install is also good, as well as [DebOps](https://github.com/debops) playbooks, many others) in a new section, separate file, ... . At the moment I'll try to work on https://github.com/Kickball/awesome-selfhosted/issues/1038, which will allow us to design more efficient ways to store/filter/display/sort/unbloat the list/make it more readable, which would in turn allow adding more software.

Let's wait for word of @Kickball who's been afk for some time

https://daringfireball.net/projects/markdown/syntax

nikksno:
No problem, thank you. I understand your point, I'm not trying to change your mind on aenigma's entry in the list, I'm simply discussing these topics out of project-related interest. Thanks again for your time and precious advice!

PR #1287, opened 2018-01-04
Add M.A.L.P., add new SPDX version 3 licenses, since M.A.L.P. uses one.

meitar:
Thank you for taking the time to work on a PR for Awesome-Selfhosted!

To ensure your PR is dealt with swiftly please check the following:

    ``- [Name](http://homepage/) - Short description, less than 250 characters. ([Demo](http://url.to/demo), [Source Code](http://url.of/source/code)) `License` `Language` ``

- [x] Your additions are ordered alphabetically.
- [x] Your additions are [Free software](https://en.wikipedia.org/wiki/Free_software), or if not they have been added to [non-free](non-free.md).
- [x] Your additions are not already listed at [awesome-sysadmin](https://github.com/n1trux/awesome-sysadmin) (IT infrastructure management), [staticgen.com](https://www.staticgen.com/) or [staticsitegenerators.net](https://staticsitegenerators.net/) (static site generators).
- [x] Any licenses you have added are in our [list of licenses](https://github.com/Kickball/awesome-selfhosted/blob/master/README.md#list-of-licenses).
- [x] You have searched the repository for any relevant [issues](https://github.com/Kickball/awesome-selfhosted/issues) or [PRs](https://github.com/Kickball/awesome-selfhosted/pulls).
- [x] Any category you are creating has the minimum requirement of 3 items.

### DIFF
--- a/README.md
+++ b/README.md
@@ -803,6 +803,7 @@ See also <https://en.wikipedia.org/wiki/List_of_streaming_media_systems>, <https
   - [Mopidy MusicBox](https://github.com/pimusicbox/mopidy-musicbox-webclient) - Web Client for Mopidy Music Server. `Apache-2.0` `HTML5`
   - [Mopidy-Party](https://github.com/Lesterpig/mopidy-party) - Mopidy web extension designed for party! Let your guests manage the sound. `Apache-2.0` `Python`
 - [mpd](http://www.musicpd.org/) - Daemon to remotely play music, stream music, handle and organize playlists. Many clients available. `GPL-2.0` `C++`
+  - [M.A.L.P.](https://f-droid.org/en/packages/org.gateshipone.malp/) - Android client for mpd with artwork/cover image download capabilities. ([Source Code](https://github.com/gateship-one/malp)) `GPL-3.0-or-later` `Java`
   - [ympd](http://www.ympd.org/) - Standalone MPD Web GUI written in C, utilizing Websockets and Bootstrap/JS. ([Source Code](https://github.com/notandy/ympd)) `GPL-2.0` `C`
 - [mStream](http://mstream.io/) - Music streaming server with GUI management tools. Runs on Mac, Windows, and Linux. ([Demo](https://darncoyotes.mstream.io/), [Source Code](https://github.com/IrosTheBeggar/mStream)) `GPL-2.0` `Nodejs`
 - [Music Browser](https://github.com/henrik242/musicbrowser) - Browser and streamer for your music collection. It is runs on most operating systems, and is light enough to run flawlessly on NAS devices. `GPL-3.0` `PHP`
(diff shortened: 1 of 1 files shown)


nodiscc:
Than you for the patch, this looks interesting (https://f-droid.org/en/packages/org.gateshipone.malp), but is this purely an android client app? We don't list clients (except those that are served though the network themselves), so this might be a better fit for https://github.com/johnjago/awesome-free-software#audio.

Let me know if you believe I closed this PR in error.

meitar:
> is this purely an android client app?

Yes.

> We don't list clients (except those that are served though the network themselves)

Oh! Okay. Hmm, that seems like a relatively arbitrary distinction to me. I thought the intention was to offer a convenient source of information about tools with which to send or receive *streamed* audio.

If I were a user browsing this list for the purpose of freeing myself from SaaSS providers such as Spotify or iTunes, I would want to find not only a good server but also, obviously, a client. If the clients listed here are purely Web-based clients (clients "served through the network themselves"), I am forced to either adapt my behavior to the tools or look elsewhere for additional clients.

From the other perspective, if the intention of awesome-selfhosted as a project is to advocate for such users who want to free themselves from SaaSS vendors and the associated vendor lock-in, then it seems counter-productive to refuse to list clients based on what seems to me to be a distinction-without-a-difference from the perspective of an end user. That is, as a project, this project becomes more useful if the most awesome clients for `mpd` (or whatever) are listed alongside with the audio streaming server itself. Not all of the clients, of course, just the *really* good ones. :)

I do not mean to usurp your intentions nor @Kickball's intentions with this list, and I may very well propose that `awesome-free-software` add M.A.L.P. to its list, as well. I'm just…curious as to the reasoning behind "we only list clients delivered through the network itself," because, if I understand correctly, this is the exact opposite of what "…locally hosting and managing applications instead of renting from SaaS providers" actually means.

By which I mean, from the perspective of maximum freedom (and, arguably, security), I far prefer to have the streaming clients I use be applications whose compute cycles happen on my device, rather than loading the client itself through the network with, for example, JavaScript and all of its problems.

Since there was no Android-based `mpd` client listed in `awesome-selfhosted`, I figured I'd share one of my favorite "awesome" ones. :)

Thanks regardless for the clarification and if you are willing to let me know why this distinction is enforced, I can provide better suggestions/PRs in the future.

nodiscc:
>  thought the intention was to offer a convenient source of information about tools with which to send or receive streamed audio.
> it seems counter-productive to refuse to list clients

I agree the distinction is arbitrary, readers of the list would be better served by a list of services + clients for these services. For example it would be nice to list CalDAV/CardDAV/XMPP/Matrix clients, and many more. However:

 * It's expected that the project homepage will give a list of available clients (too many times they don't)
 * Our current major problem is to keep the list readable (as you can see it's already a bit overloaded, and this has been a recurrent complaint/feedback: "not enough curation"), so extending the scope is not going to help. The only way I see non-web clients getting into the list is to work on a machine-readable list storage/database backend (https://github.com/Kickball/awesome-selfhosted/issues/1038), add these clients to the list, and render the list in a way that related software/clients do not clutter the "default" view of the listing.

Feel free to check the mentioned issue and/or add your suggestions for client list formatting at https://github.com/nodiscc/awesome-selfhosted-yaml (currently on hold as noone seems to have the time to work on it). There was some discussion in https://github.com/nodiscc/awesome-selfhosted-yaml/pull/1.

meitar:
> It's expected that the project homepage will give a list of available clients (too many times they don't)

As a free software developer, this seems like an unreasonable expectation to have of free software server projects, to me. If I werw writing a CalDAV server, for instance, I would not want to be tasked with monitoring the Web for *other people's projects* that interoperated with mine and then cataloguing those other projects. That would be exactly what I'd expect that a list like this would/should be for.

> For example it would be nice to list CalDAV/CardDAV/XMPP/Matrix clients, and many more.

"Would be nice?" Wait, now I'm confused. We *do* list CardDAV/CalDAV clients. See, for instance, DAVDroid, which is an Android-only CalDAV/CardDAV client. Why does that client make sense to add to this list while M.A.L.P. does not? Both are Android-only clients and both receive data over network streams of one protocol or another. By adding M.A.L.P., I am simply trying to make the Audio streaming section as useful as the Calendaring section, which is one of several sections that seem to already the violate the arbitrary "browser-only clients" rule?

nodiscc:
DAVDroid was added in error, it should be removed

meitar:
> DAVDroid was added in error, it should be removed.

Oh. Hmm. I guess I really don't understand the intent of this project, then. Thanks for all your hard work on the foundation, so far! :)

PR #1292, opened 2018-01-10
Piwik is now called Matomo

kumitterer:
The project formerly known as Piwik has now changed its name to Matomo

PR #1301, opened 2018-01-15
Rename Piwik to Matomo

Lanseuo:
Piwik was renamed to Matomo: https://matomo.org/blog/2018/01/piwik-is-now-matomo/


nodiscc:
Duplicate of https://github.com/Kickball/awesome-selfhosted/pull/1292

PR #1306, opened 2018-02-01
Remove my name + email address from AUTHORS.md

cpdevelops:
Hello there. A while ago, I contributed some formatting changes, additions, and other issues to this project. Those commits added my name and multiple email addresses to the AUTHORS.md file. I would like those removed as I am now getting various spam emails because of it. As well as other people from Github, probably people on this list (see screenshot) If this commit could be merged that would be great.

This is the seventh spam email I have gotten in the past 5 days, and this is the only place my name is on the internet and as you can see there are other Github email addresses sent this calendar inv.
image

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -37,8 +37,6 @@ Commits | Author
      4 | Alexandr Emelin <frvzmb@gmail.com>
      4 | AndrewCz <smacz42@users.noreply.github.com>
      4 | Chris McCormick <chris@mccormick.cx>
-     4 | Colin Pokowitz <colin@cpdev.me>
-     4 | Colin Pokowitz <colinpokowitz03@gmail.com>
      4 | Cory Gibbons <hello@corygibbons.com>
      4 | D <DL88250@gmail.com>
      4 | Dominik Pfaffenbauer <dominik@lineofcode.at>


evaryont:
I'm sorry you're getting spam, though I'd like to make sure you realize this doesn't completely remove your address. It's still present in previous revisions, and it's still associated with the git commits. Not to mention projects like Internet Archive and similar.

It will make it harder for bots to scrape it, so there's still value in going ahead with this.

cpdevelops:
Yes, I understand that this will not make it impossible, but it will hopefully dampen it.

nodiscc:
I have merged this, but keep in mind that this file is automatically generated from the git log, so it will eventually reappear. A permanent fix is to edit https://github.com/Kickball/awesome-selfhosted/blob/master/.github/.mailmap so that another address is displayed in AUTHORS.md.

A fix for future commits is to set `user.email` to a spam/github only inbox address in your git configuration (`git config --global user.email`). Some mail providers allow setting a mail alias such as `myreealemail+spam@gmail.com`.

As @evaryont said the git log and previous revisions will always retain your address. `user.email` is part of how `git` works, it is public information but a plain text AUTHORS.md probably worsened your problem. We could add a reminder in CONTRIBUTING.md to change `user.email`/`.mailmap` for contributors who don't want their address to appear in AUTHORS.md.

Additionally it would be interesting to document a full/efficient antispam setup for self-hosted email servers :))) I admit that managing mail without it is next to impossible (so much spam... about [90%](https://en.wikipedia.org/wiki/Email_spam) according to Wikipedia). I have left my Github address on Gmail for that reason, but will reconsider it if I can find a way to manage spam.

PR #1322, opened 2018-02-26
Fair Source License is not a FLOSS license

disk0x:
according to .github/CONTRIBUTING.md, the main list must only contain Free Software.
Fair Source License does not meet any authoritative definition of FLOSS (FSF, Debian, OSI)
In their words: "The Fair Source License is not an open-source license and doesn’t intend to be an open-source license."


nodiscc:
Thanks. It was only used for Buddy Go which was also removed recently: https://github.com/Kickball/awesome-selfhosted/pull/1314

PR #1373, opened 2018-04-21
remove Mumblecop

nodiscc:
A comprehensive list of mumble bots can be found at https://wiki.mumble.info/wiki/3rd_Party_Applications

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,7 +259,6 @@ _CalDAV or CardDAV clients._
 - [MiAOU](https://dystroy.org/miaou/login) - Multi-room persistent chat server. ([Source Code](https://github.com/Canop/miaou)) `MIT` `Nodejs`
 - [Mibew](https://mibew.org) - Mibew Messenger is an open-source live support application written in PHP and MySQL. It enables one-on-one chat assistance in real-time directly from your website. ([Demo](https://mibew.org/demo2), [Source Code](https://github.com/Mibew/mibew)) `Apache-2.0` `PHP`
 - [Mumble](http://wiki.mumble.info/wiki/Main_Page) - Low-latency, high quality voice/text chat software. ([Source Code](https://github.com/mumble-voip/mumble)) `BSD` `C++`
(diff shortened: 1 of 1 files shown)


evaryont:
And I have confirmed that Mumblecop is on that wiki page.

Do we want to link to that page? If we're going to start including client software, this seems like a good place to start. A reference along the lines of "there are plenty of open source clients and bots you can host yourself for Mumble. See their wiki for more information" as a sub-bullet?

nodiscc:
Should we allow another link _Clients_ in addition to the _Demo_ and _Source code_ links? The mumble link would look like


- [Mumble](http://wiki.mumble.info/wiki/Main_Page) - Low-latency, high quality voice/text chat software. ([Source Code](https://github.com/mumble-voip/mumble), [Clients](https://wiki.mumble.info/wiki/3rd_Party_Applications)) `BSD` `C++`

Same could probably be done for other software (mpd and openstreetmap come to mind, ping @meitar, as well as icecast - https://github.com/Kickball/awesome-selfhosted/pull/1359). That way we could point to the projects' official client listings, or a community-contributed list of clients/pass (eg. https://github.com/osmlab/awesome-openstreetmap)

evaryont:
For projects that maintain their own list of clients (or community members that do that on a project's behalf) it does make sense to me to include a *Clients* link. I think I've come around to agree with @meitar that we should include links to clients, when projects don't maintain their own list.

nodiscc:
@evaryont I agree that it would be useful.

> https://github.com/Kickball/awesome-selfhosted/pull/1287: If I were a user browsing this list for the purpose of freeing myself from SaaSS providers such as Spotify or iTunes, I would want to find not only a good server but also, obviously, a client.

My only concern is not to make the list more confusing than it already is. I think I have found a compromise that would allow both. I will propose a solution in this PR once https://github.com/Kickball/awesome-selfhosted/pull/1385 is merged (consolidating documentation for contributors).

Can you or someone review it please? (being the only active maintainer has become a problem I think, , thanks for your help so far)

arylatt:
We could link to it as a clients section (we may need to update travis/danger files), would it perhaps work listing them under External Links? The downside of that being that it separates the clients quite significantly from the original listing...

nodiscc:
PR obsoleted by https://github.com/Kickball/awesome-selfhosted/pull/1410

PR #1374, opened 2018-04-23
Fix git-annex url

karloluiten:
Simple URL fix for git-annex.


Kickball:
@karloluiten @evaryont, I've come to this late due to some medium-term issues so apologies.

This PR appears to change the homepage from what looks to be a homepage to a copy of the Github page on a publically accessible file system. Was the situation the same when this PR was put through? If so, why was this change made?

On a side note the Github repo linked looks to have gone down. #1401 raised for this.

PR #1369, opened 2018-04-19
Added Ambar to open-source solutions

sochix:
It's our second submission to Awesome Selfhosted with Ambar. Now Ambar fully open-source under MIT license

- [x] Your submissions are formatted according to the guidelines. 
        
    ``- [Name](http://homepage/) - Short description, less than 250 characters. ([Demo](http://url.to/demo), [Source Code](http://url.of/source/code)) `License` `Language` ``

- [x] Your additions are ordered alphabetically.
- [x] Your additions are not already listed at [awesome-sysadmin](https://github.com/n1trux/awesome-sysadmin) (IT infrastructure management), [staticgen.com](https://www.staticgen.com/) or [staticsitegenerators.net](https://staticsitegenerators.net/) (static site generators).
- [x] Your additions are [Free software](https://en.wikipedia.org/wiki/Free_software), or if not they have been added to [non-free](non-free.md).
- [x] Any licenses you have added are in our [list of licenses](https://github.com/Kickball/awesome-selfhosted/blob/master/README.md#list-of-licenses).
- [x] You have searched the repository for any relevant [issues](https://github.com/Kickball/awesome-selfhosted/issues) or [PRs](https://github.com/Kickball/awesome-selfhosted/pulls).
- [x] Any category you are creating has the minimum requirement of 3 items.
- [x] Any software project you are adding to the list is actively maintained.

PR #1400, opened 2018-06-13
Add talkyard

Kickball:
Add talkyard to the Social Networks and Forums section.

Thank you for taking the time to work on a PR for Awesome-Selfhosted!

To ensure your PR is dealt with swiftly please check the following:

    ``- [Name](http://homepage/) - Short description, less than 250 characters. ([Demo](http://url.to/demo), [Source Code](http://url.of/source/code)) `License` `Language` ``

- [x] Your additions are ordered alphabetically.
- [x] Your additions are [Free software](https://en.wikipedia.org/wiki/Free_software), or if not they have been added to [non-free](non-free.md).
- [x] Your additions are not already listed at [awesome-sysadmin](https://github.com/n1trux/awesome-sysadmin) (IT infrastructure management), [staticgen.com](https://www.staticgen.com/) or [staticsitegenerators.net](https://staticsitegenerators.net/) (static site generators).
- [x] Any licenses you have added are in our [list of licenses](https://github.com/Kickball/awesome-selfhosted/blob/master/README.md#list-of-licenses).
- [x] You have searched the repository for any relevant [issues](https://github.com/Kickball/awesome-selfhosted/issues) or [PRs](https://github.com/Kickball/awesome-selfhosted/pulls).
- [x] Any category you are creating has the minimum requirement of 3 items.


Kickball:
Please wait for the response on [#27](https://github.com/debiki/talkyard/issues/27) to understand whether there is a dependency on Docker before merging.

Kickball:
[#27](https://github.com/debiki/talkyard/issues/27) has confirmed that the project relies on Docker.

@nodiscc how have we previously documented this, do we add a ⚠ or list Docker as a language?

I think the first option makes the most sense.

evaryont:
Docker, to me, seems similar to specifying Linux for most of these links. It's simply a convenient wrapper around the container technologies that have existed in the Linux kernel for years already. Nor is it non-free, it's [open source](https://github.com/moby/moby), though open-core, with layers on top that are proprietary for the Enterprise market.

arylatt:
There doesn't seem to be any consistency as to whether or not Docker should be listed as a language. Some items do have it and some items don't have it. Either we can standardize having or not having it listed, or we can leave it up to the submitter. If we're opting for the second approach then this is fine to merge.

kajmagnus:
Maybe it'd be interesting to know, that there's software on the Awesome-Selfhosted list already, that requires Docker — I have in mind Discourse:
https://github.com/discourse/discourse/blob/master/docs/INSTALL.md ("Why do you only officially support Docker?")

Talkyar is similar — only Docker supported, ... and if someone really wants to, s/he can install without Docker and then s/he needs to figure out how, themselves.

People have done that, with Discourse (figured out themselves how to install Discourse without Docker), for example:  https://hackernoon.com/how-i-ran-an-unsupported-discourse-install-with-nanobox-f6736f83cef0  — and not with Talkyard (which is younger than Discourse, doesn't have that many users).

From that HackerNoon blog post, about Discourse:

> I don’t call this out *[i.e. only supporting Docker]* because I think there is anything wrong with what they’re doing. In fact the exact opposite. I think it makes complete sense. How could a small team be expected to support so many possible installs of their project?

nodiscc:
Hi everyone, this is the way I use to handle software for which only a docker-based installation method is supported:

 * Must have `Docker` language/platform tag. The main language should also be listed, so for Talkyard `Scala/Docker`
 * A way to install from source without Docker (even undocumented) should be available to the user, else it goes in non-free.md. This means all the docker images used in the build process must be available for inspection/modification. For talkyard the install process (https://github.com/debiki/talkyard-prod-one) uses a few [scripts](https://github.com/debiki/talkyard-prod-one/tree/master/scripts) that prepare an Ubuntu machine (not ideal for me, I wouldn't want an application/service install to touch my system's configuration - just my 2 cents, but I realize it's easier to support known environments for the devs) and this [docker-compose](https://github.com/debiki/talkyard-prod-one/blob/master/docker-compose.yml) file, which downloads these docker images [[1]](https://hub.docker.com/r/debiki/talkyard-app/) [[2]](https://hub.docker.com/r/debiki/talkyard-web/) from Docker Hub, no Dockerfile is available for these images, nor can I find one in https://github.com/debiki?tab=repositories.

Hence it would be added to non-free until the full source/build process is published.

See https://github.com/Kickball/awesome-selfhosted/pull/1369 for a similar discussion

kajmagnus:
@nodiscc The Dockerfiles are in the `talkyard` repo, located at `docker/<image-name>/Dockerfile`, for example, for Nginx (the *web* image):

https://github.com/debiki/talkyard/blob/master/docker/web/Dockerfile

The build script:  https://github.com/debiki/talkyard/blob/master/s/build-and-release.sh

I realize I need to improve the docs so people find them. I can 1) add a comment in the docker-compose file, about where to find the Dockerfiles, and 2) mention in the `talkyard-prod-one` readme that the source code & build script etc are in the `talkyard` repo, and 3) I suppose I can somehow add a comment in https://hub.docker.com/r/debiki/talkyard-app/ (& other images) that links to the Dockerfiles.

(If there's something else that seems missing — it *is* there somewhere, it's just that I didn't think about adding docs about it.  & If you can think of more things that people might go looking for, and that I could document, it'd be interesting to know :- ))

kajmagnus:
@nodiscc about: *"I wouldn't want an application/service install to touch my system's configuration"* —  I also would want the insall to avoid that.  ... ElasticSearch needs one to change a `/etc/sysctl.conf` conf val, and there're cron jobs for backups & auto upgrades. Any thoughts / ideas about less intrusive ways to get those things done, would be interesting to hear. Or if maybe there're any standard tools you know about ...

... Someone suggested a tool named [Watchtower](https://github.com/v2tec/watchtower) for auto-upgrades — not a good fit in this case though (e.g. cannot upgrade all images at the same time, and no undo / rollback).  ... Maybe backups & upgrades could be handled by background threads & timeouts inside the containers, instead of cron jobs. On the other hand, maybe sysadmins like to get an overview by typing `crontab -l`, hmm.

nodiscc:
@kajmagnus 1) 2) 3) Yes. You can also setup [Automated builds](https://docs.docker.com/docker-hub/builds/) and make the Dockerfile public there.

> ElasticSearch needs one to change a /etc/sysctl.conf conf val, and there're cron jobs for backups & auto upgrades

I advise you put the preliminary requirements in the documentation (packages, sysctl tuning, backups...), and don't attempt to automate system configuration from a script. You can use the script as an example in the docs, noting that it is for an Ubuntu 1x.xx machine, etc. Following a well-documented setup procedure is MUCH better than blindly running a script, it allows for user customizations, running it on other platforms... Administrators probably also have their own backup tools they want to use, etc. You could script some checks such as the sysctl parameter, just warning the user and returning an error code and message.

> Someone suggested a tool named Watchtower for auto-upgrades

Again leave it to distribution tools like [unattended-upgrades](https://wiki.debian.org/UnattendedUpgrades) for things that are outside the application's responsibility, and document the upgrade procedure. For the application itself of course you can write your own updater.

As a last step if you really want to automate the whole setup process you could setup another repository featuring automation tools like [Ansible](https://www.ansible.com/) which are less error-prone than shell scripts. You can check some examples in the roles directory [here](https://github.com/nodiscc/srv01) (if you look at the very beginning of the git history this tool used to be a bunch of bash scripts... it was not pretty) or there are ansible howtos all over the web.

I hope it helps.

If it can be completely built from source, good to merge

Issue #1415, opened 2018-06-18
wrong section for Jappix resp. a mess in conjunction with Movim


DJCrashdummy:
somethings wrong with `Line 414`:

> - [Jappix](https://jappix.com/) - Jappix is an open social platform, that let's you easily get or keep in touch with everyone. ([Source Code](https://github.com/jappix/jappix)) `AGPL-3.0` `PHP`

- first it is listed in the section _"Social Networks and Forums"_, but as far as i know **Jappix** is a XMPP client (have a look the source link).
- and the homepage-link leads to a movim.eu login-page.

and after having a look at _"XMPP Web Clients"_ it gets finally confusing at `Line 456`:

> - [Movim](https://movim.eu/) - Modern, federated social network based on XMPP, with a fully featured group-chat, subscriptions and microblogging. ([Source Code](https://github.com/movim/movim)) `AGPL-3.0` `PHP`

- although it is based on XMPP and somehow related to Jappix, i guess **Movim** fits better at _"Social Networks and Forums"_... at least definitely much more than Jappix!

btw: i'm not sure whether it is worth to still list **Jappix** at _"XMPP Web Clients"_ resp. at all, since it is stated at its source: `"the Jappix project is no longer maintained. We still accept PR (Pull Requests) though."`

Kickball:
I'm not certain on the best section for Movim as I'm familiar with XMPP, but I'm leaning towards moving it to the _Social Networks and Forums_ section as it has features outside of XMPP (blogging/Facebook-like timeline).

@nodiscc your view on the above would be appreciated if you can.

Jappix has not been updated for 18 months and is not maintained as you stated in the issue, I've therefore raised #1422 to remove it from the list.

nodiscc:
Merged #1422, I think @DJCrashdummy is right and Movim should be moved to _Social Networks and Forums_

PR #1424, opened 2018-06-20
Move Movim to Social Networks & Forums

Kickball:
Move Movim from XMPP web clients section to Social Networks & Forums section.


Kickball:
Raised to resolve #1415.

PR #1498, opened 2018-09-19
Add FreeScout

freescout-helpdesk:
Added FreeScout ticketing app.

PR #1499, opened 2018-09-19
Added FreeScout

freescout-helpdesk:
Added FreeScout ticketing app.


Kickball:
#1498 was the previous, related PR.

Issue #1234, opened 2017-11-09
Add Airsonic


nodiscc:
https://github.com/airsonic/airsonic
https://airsonic.github.io/
Free and Open Source community driven media server (fork of Subsonic and Libresonic)
Java/GPLv3

PR #1509, opened 2018-09-24
Replace Libresonic with Airsonic

jooola:
Hi, I would like to replace Libresonic with Airsonic since we forked again more than a year ago. Reason why can be read [here](https://airsonic.github.io/#history), also Libresonic is dead since we forked.


Kickball:
Fixes #1234.

PR #1520, opened 2018-10-03
Add LocalStack to FaaS

jciskey:
LocalStack provides a local cloud environment that matches AWS APIs and functionality for included components. This includes Lambda functions for serverless computing.


jciskey:
I can easily make that change, but the reason I went with Multiple instead of Apache 2.0 is that the third-party software they build on top of all have other licenses.

With that in mind, did you still want me to make the change?

n8225:
@nodiscc what are your thoughts?

n8225:
After further review of a few apps already on the list, I believe it should be listed as Apache-2.0  Many apps already on the list have dependencies with different licenses. 

@nodiscc @Kickball @arylatt  any thoughts on this?

PR #1563, opened 2018-11-13
Add Thredded Forums

glebm:
Not sure if it fits this list, as the setup is a bit more involved.

Thredded has a Rails app generator and a set of Ansible playbooks for provisioning and deployment.


glebm:
Done

PR #1606, opened 2019-01-12
Add trillium

cooltrooper:
Added trillium, an interesting local note library with self hosted option


nodiscc:
@cooltrooper Can you swap order of `Nodejs` and `AGPL-3.0`? Thanks

PR #1611, opened 2019-01-20
ASTPP

Mancy:
Open Source VoIP Billing Solution for Freeswitch

### DIFF
--- a/README.md
+++ b/README.md
@@ -392,7 +392,6 @@ _[IRC](https://en.wikipedia.org/wiki/Internet_Relay_Chat) communication software
 _[SIP](https://en.wikipedia.org/wiki/Session_Initiation_Protocol) telephony software_
 
 - [Asterisk](http://www.asterisk.org/) - Easy to use but advanced IP PBX system, VoIP gateway and conference server. `GPL-2.0` `C`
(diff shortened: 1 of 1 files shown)


Mancy:
@nodiscc As i can see, this is not the case with Freepbx and FusionPBX
maybe ASTPP should be moved to the IPBX section instead?

nodiscc:
@Mancy you are right, basically Asterisk and FreeSWITCH should be moved to IPBX section and this would be a good way to list software based on them:

#### IPBX

- Asterisk
  - Freepbx
- FreeSWITCH
  - ASTPP
  - FusionPBX

Can you move the items as I suggested?

nodiscc:
ping @Mancy

Mancy:
@nodiscc 
I have just moved the ASTPP to IPBX category since both asterisk and Freeswitch are SIP, and thus i am confused to move them to IPBX

nodiscc:
@Mancy You are right... Now I'm wondering about merging these sections into a single `SIP/IPBX` section. There is overlap between all of these projects...

Mancy:
Alright I will merge the two sections, as a non-nested list, since it's a bit complicated to nest the list, for example Kamailio sometimes used with Asterisk or FreeSwitch

PR #1634, opened 2019-02-24
Added Fathom PRO to list

JackEllis:
Fathom is open source code, but users can also pay for a managed service!


nodiscc:
Since the non-Free Software version cannot be hosted on your own server I see no reason to add it to the list.

nodiscc:
@JackEllis feel free to ping us if you release a Pro version with the ability to self-host it.

PR #1699, opened 2019-05-20
Update link and license for ShinobiCE

PrplHaz4:
Code moved to gitlab, license has changed from GPL-2.0 to GPL-3.0 + AGPL-3.0


evaryont:
To me, that licensing doesn't make sense. Isn't AGPL 3 a superset if GPL 3? Oh well, I'm not a lawyer.

Nit: Could you change the license description to be `AGPL-3+GPL-3`? This way there is only 2 inline code blocks at the end, making it easier to machine parse.

PrplHaz4:
@evaryont nor am I, but I think it has something to do with a commercial version of the product also being available.

That is how I had the tag initially, but changed it to de individual tokens - will update ASAP!

README.md
```diff
@@ -892,7 +892,7 @@ See also <https://en.wikipedia.org/wiki/List_of_streaming_media_systems>, <https
 - [Odd Networks](http://oddnetworks.com) - Open source video management system (VMS) and API for collections and videos with supporting SDKs for Roku, Apple iOS/tvOS, Google Android, and Amazon FireTV. ([Source Code](https://github.com/oddnetworks/oddworks)) `MIT` `Nodejs`
 - [PeerTube](https://joinpeertube.org/en/) - Decentralized video streaming platform using P2P (BitTorrent) directly in the web browser. ([Source Code](https://github.com/Chocobozzz/PeerTube)) `AGPL-3.0` `Nodejs`
 - [Restreamer](https://datarhei.github.io/restreamer/) - Restreamer allows you to do h.264 real-time video streaming on your website without a streaming provider. ([Source Code](https://github.com/datarhei/restreamer)) `Apache-2.0` `Nodejs/Docker`
-- [Shinobi](https://moeiscool.github.io/Shinobi/) - Open Source CCTV software written in Node with both IP and local camera support. ([Source Code](https://github.com/moeiscool/Shinobi)) `GPL-2.0` `Nodejs`
+- [ShinobiCE](https://gitlab.com/Shinobi-Systems/ShinobiCE) - Open Source CCTV software written in Node with both IP and local camera support. ([Source Code](https://gitlab.com/Shinobi-Systems/ShinobiCE) `AGPL-3+GPL-3` `Nodejs`
```
nodiscc: - The "correct" character to separate multiple licenses or languages is a slash.  
- The [SPDX license idientifiers](https://spdx.org/licenses/AGPL-3.0.html) are AGPL-3.0 and GPL-3.0.  
- So the correct notation would be `AGPL-3.0/GPL-3.0`.  
- I think relicensing GPLv3 to AGPLv3 requires formal agreement from all previous contributors, so it makes sense that the project is dual-licensed until this happens.

evaryont: Ah, thanks for the correcting the syntax on the license.

I didn't realize they were in the process of changing license, that would explain the duality.

PrplHaz4: @nodiscc @evaryont updated PR with correct syntax

PR #1725, opened 2019-06-24
Add elixire

aveao:
Went through checklist.

This adds our FOSS image hosting thing, elixire.

I linked to the gitlab org, which contains backend, frontend, CLI client, api docs and more (just running backend is enough to get an elixire instance running, though one will have to use API to interact with it).

Dunno if the website link is correct. Should I copy the gitlab org to that link and move the service link to the demo link?


aveao:
Done!

PR #1740, opened 2019-07-11
Add Square to Money category

bobbilee19:
https://developer.squareup.com/docs/

A couple of Square SDK examples...
Java - https://github.com/square/connect-java-sdk
Python - https://github.com/square/connect-php-sdk
Ruby - https://github.com/square/connect-ruby-sdk/


evaryont:
This is not self-hosted at all. Where is the code to run your own payment processor? The SDKs are just ways to connect to Square; not a way to run your own competitor/alternative to Square.

Most of the software you'll find on this list is the backend for such SDKs to be developed for.

nodiscc:
Confirming what was said above.

Also we don't list SDKs/libraries/application building blocks.

PR #1758, opened 2019-07-28
Edit: Better description and right name for Riot.im

jtagcat:
It's Riot.im, not "RIOT".
Also the description felt off so I grabbed the official description.


jtagcat:
I should probably add everything from https://matrix.to

PR #1798, opened 2019-08-30
Update README.md

peponi:
add **dockerbunker** category **Self-hosting Solutions**


nodiscc:
> https://github.com/chaosbunker/dockerbunker#why-i-made-this
> I know that it is not really ideal and recommended to do something like this with shell scripts. dockerbunker is an idea that went a bit out of control. 

:) Please check the [shellcheck output](https://pastebin.com/KYxvsqMU). I went the pure shell script way before with similar projects, it takes some work to make it right. I moved to configuration management.

The CLI/interactive wrapper is nice, but the project uses shell scripts with `docker run`, why not use [docker services](https://docs.docker.com/get-started/part3/) for this?

It seems to install [`latest`](https://github.com/chaosbunker/dockerbunker/blob/master/data/services/gitlabce/gitlabce.sh#L20) tags for Docker images. How does it manage major/breaking version upgrades/what is the process to get the latest security upgrades?

nodiscc:
No reply for 15 days, closing. Feel free to reopen if more info can be provided.

PR #1880, opened 2019-11-12
Update demo link for Wallabag

phobot:
The link to http://demo.wallabag.org/ is no longer working, currently pointing to an nginx welcome page. New link should be https://app.wallabag.it/register/


StephanMa:
New link results in 502 server error

SuperSandro2000:
> New link results in 502 server error

works for me

gregf:
It was 502 earlier, main page is just a standard nginx install page though.

PR #1930, opened 2019-12-30
Update README.md

syedrali:
Added entry for Ubooquity book server (https://vaemendis.net/ubooquity/) under E-books and Integrated Library Systems.


syedrali:
Software is free to use but source code isn't licensed under an open-source license so removing entry.

PR #1946, opened 2020-01-14
Remove linkr

SuperSandro2000:
The project is not awesome as it does not accept any issues at all. Also there is no longer a demo or any screenshots.


evaryont:
I'm not sure if that's a criteria I can get behind. Just because you release your software's source code, doesn't require you listen to other people. It might be better, but it's not mandated. To me, to require that would just mean the issues in Github are unmoderated. Unmoderated issues is not better than inaccessible/0 issues.

SuperSandro2000:
The problem I have right now is that you can't contact him in anyway except searching for him on the web or hijacking a PR or another project.
I don't believe this is required for a project to be open source but it does not sound awesome.

nodiscc:
Agree with @evaryont , thanks for the suggestion anyway.

PR #1959, opened 2020-01-29
Update regex pattern to allow Clients.

### LINKED ISSUE #1956: Tests: an entry with only a Clients link (no source code/demo links) should be considered valid
Ref https://github.com/awesome-selfhosted/awesome-selfhosted/pull/1954

n8225:
This will fix #1956.  It will allow multiple Clients, but it enforces the order from the PR template (Demo, Source Code, Clients).  I could easily change it to allow any name for the client (web, android, ios, etc).
(It also found about 10 more items with minor syntax errors.)

### DIFF
--- a/tests/test.js
+++ b/tests/test.js
@@ -16,32 +16,11 @@ function split(text) { // Function to split lines into array
   return text.split(/\r?\n/);
 }
 
-function findPattern(text) { // Test entries against 8 patterns.  If matches pattern returns true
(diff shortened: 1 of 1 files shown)


nodiscc:
> I could easily change it to allow any name for the client

I'd prefer to leave it as is, the goal is to keep a single CLients link linking to a well maintained list of clients. Thanks!

PR #1979, opened 2020-02-08
Fix Multiple Syntax Errors

n8225:
Fixes syntax on:
Kiwi IRC
indico
Teedy
Calibre
FreshRSS
Hospital Run
LimeSurvey
MantisBT
Shlink
Mediawiki


nodiscc:
Thank you

PR #2001, opened 2020-02-26
Add notice under link shorteners to prevent link rot.

jtagcat:
TinyURL, bit.ly, and other similar services allow long URLs to be converted to smaller ones on their specific service; the small URL is visited by a consumer and their web browser is redirected to the long URL.

Such services are a ticking timebomb. If they go away, get hacked, or sell out, millions of links will be lost (see [Wikipedia: Link Rot](https://en.wikipedia.org/wiki/Link_rot)). [Archive.org/301Works](https://archive.org/details/301works) is acting as an escrow for URL shortener databases, but they rely on URL shorteners to actually give them their databases. Even 301Works founding member bit.ly does not actually share their databases and most other big shorteners don't share theirs either. 

- [x] You have searched the repository for any relevant [issues](https://github.com/awesome-selfhosted/awesome-selfhosted/issues) or [PRs](https://github.com/awesome-selfhosted/awesome-selfhosted/pulls), including closed ones.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1444,8 +1444,7 @@ See https://staticsitegenerators.net and https://www.staticgen.com
 
 ## URL Shorteners
 
-URL Shorteners are a ticking time bomb. If they go away, get hacked, or sell out, millions of links will be lost. 
-If you decide to host one, please add it to [Archive Team](https://www.archiveteam.org/)'s [URLTeam](https://www.archiveteam.org/index.php?title=URLTeam) list of Alive URLs.
+Before hosting one, please see [shortcomings](https://en.wikipedia.org/wiki/URL_shortening#Shortcomings) of URL shorteners.
 
 **[`^        back to top        ^`](#)**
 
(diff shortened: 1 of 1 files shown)


nogweii:
Hm, I appreciate the concern. But aren't some of those concerns the reason why you'd want to self-host a URL shortner in the first place? That way you have control when the domain expires, or if the URL it's linking to is against policy. You could have less control when it's managed by a 3rd party.

In a similar vein, most of these applications can be vectors for spam and abuse, especially if people don't set up authentication. How many file sharing webapps on this list allow anonymous uploads? Pastebins without requiring authentication? Generally, by being a sysadmin of a self-hosted tech, you already must do work to prevent that spam/abuse. Why would URL shortners need to be specifically called out opposed to any other of these?

Maybe such language warning about the responsibility you take on could be added towards the bottom of the document? (Maybe "External Links", but that doesn't fit super well. Perhaps a new section about warnings or recommendations.)

nodiscc:
I'd also rather take some time to write a short disclaimer/warning/general recommendation section about responsibililty, availability/backups/redundancy, link rot, security concerns,... when self-hosting services - than adding a notice specifically for URL shorteners (these problems about spam/link rot are not specific to shorteners)

PR #2068, opened 2020-04-23
Note Taking: Add GitJournal

vHanda:
GitJournal is a mobile first notes application. It stores its notes in markdown and uses Git to sync the notes.


fbartels:
While the git repo in the end can be self-hosted the software you are linking to is an app, that needs to be compiled (or downloaded from an app store) and installed on a smartphone.

vHanda:
You're correct. It doesn't have anything to host so far. Not until I finish the web version.

I figured since Joplin also does not have any server side code, the pull request would be fine.

Shall I just close this?

nodiscc:
Yes it should be closed, https://github.com/GitJournal/GitJournal -> mobile (android/ios) application hence does not qualify for the list. Unfortunately can't be found on https://f-droid.org/en/packages/

PR #2224, opened 2020-09-15
fix indentation for rss2twtxt

thomasleveil:


nodiscc (commented):
The indentation is correct, these are sub-items of Tiny Tiny RSS (like rss2twtxt is a sub-item of twtxt.net)

nodiscc (changes requested):
Oh sorry I didn't see the full change. Tiny Tiny RSS sub-items indentation was correct (2 spaces), but rss2twtxt was not (3 spaces). Please change it to 2 spaces

nodiscc (approved):

PR #2236, opened 2020-09-24
Remove IG Monitoring

philipp-r:
IG Monitoring will no longer be developed or actively supported and should be removed from the list. 
see https://github.com/jakim/ig-monitoring/issues/138


nodiscc (approved):
Thanks

PR #2244, opened 2020-10-01
Change My E-Mail Address

ProfessorLogout:


nodiscc:
This won't work. The file is auto-generated from git committer names in the repo history (https://github.com/awesome-selfhosted/awesome-selfhosted/blob/master/Makefile#L32). The next time we run `make authors` it will overwrite this change.

PR #2265, opened 2020-10-07
Updated Boost Note name and source code URL

xpac1985:
Just updated the name (Boostnote -> Boost Note) and source code URL as they released (yet another) complete rework of their app.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1113,7 +1113,7 @@ See https://github.com/n1trux/awesome-sysadmin#monitoring and https://github.com
 *See also [Wikis](#wikis)*
 
 - [BulletNotes](https://bulletnotes.io/) - Workflowy / Dynalist clone with Kanban (Trello) and Calendar functionality. Organize everything. ([Source Code](https://gitlab.com/NickBusey/BulletNotes)) `MIT` `Nodejs`
-- [Boostnote](https://boostnote.io/) - The note-taking app for programmers that focuses on markdown, snippets, and customizability. ([Source Code](https://github.com/BoostIO/Boostnote)) `GPL-3.0` `JavaScript`
+- [Boost Note](https://boostnote.io/) - The note-taking app for programmers that focuses on markdown, snippets, and customizability. ([Source Code](https://github.com/BoostIO/Boostnote.next)) `GPL-3.0` `JavaScript`
 - [CodiMD](https://demo.codimd.org/) - Realtime collaborative markdown notes on all platforms, formerly HackMD CE. ([Source Code](https://github.com/hackmdio/codimd)) `AGPL-3.0` `Node.js`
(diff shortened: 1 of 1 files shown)


nodiscc (commented):
https://github.com/BoostIO/Boostnote make no mention of https://github.com/BoostIO/Boostnote.next being its continuation. There are 3 different releases pages: [1](https://github.com/BoostIO/BoostNote.next/releases), [2](https://github.com/BoostIO/Boostnote/releases), [3](https://github.com/BoostIO/boost-releases/releases/). 

Maybe the original project README should be updated to redirect to .next, and the original project should be archived? Unless I misunderstand

.

xpac1985:
It's actually right at the top of the old readme.md ;-)
https://github.com/BoostIO/Boostnote/blob/master/readme.md

> We've launched desktop and mobile app of the new Boost Note now.

And that text links right to the new repo.

PS: I'm not in any way involved into the app, I just noticed it.

n8225:
Is this limited to local storage only with out signing into Boost Hub?

 I don't see this as being anything more than a web client or desktop app.  I'm not so sure this is a selfhosted app and I think it should be removed from the list.

xpac1985:
I'm using it without having signed in anywhere, local storage is 100% possible (I'm syncing via Seafile).

It's the usual Electron crap (custom chrome wrapped in .exe), but it's working offline, stores to local disk. 

I'd agree it's not technically hosted (as to my understanding), as you don't setup your own server though (you just use it like you'd use e.g. Notepad++ or vim), but I think that's now true for a few apps on this list?

n8225 (approved):

nodiscc:
> I'd agree it's not technically hosted (as to my understanding), as you don't setup your own server though (you just use it like you'd use e.g. Notepad++ or vim)

We had this case on several occasions (application saves data to local file -> you can sync it using self-hosted file sync -> hence it is self-hosted), and I think this reasoning is flawed. That would make LibreOffice or Keepass "self-hosted", even if these apps cannot be accessed direclty through the network.

I remember Pull Requests for such cases were rejected (don't know exactly which ones, probably somewhere in https://github.com/awesome-selfhosted/awesome-selfhosted/pulls?q=is%3Apr+label%3Ainvalid+is%3Aclosed). So I'd tend to reject this one also...

> that's now true for a few apps on this list?

Depending on the exact case I think that's a bug... which ones are you thinking about in particular?

xpac1985:
> We had this case on several occasions (application saves data to local file -> you can sync it using self-hosted file sync -> hence it is self-hosted), and I think this reasoning is flawed. That would make LibreOffice or Keepass "self-hosted", even if these apps cannot be accessed direclty through the network.

I totally understand this and have no emotions here, I just wanted to fix the link. To be fair, that would mean removing Boostnote altogether because the old version would fall under the same reasoning. It can save to disk. No hosting involved. ;)

nodiscc:
Thanks for pointing this out. I opened https://github.com/awesome-selfhosted/awesome-selfhosted/issues/2303 to remove boostnote.

PR #2305, opened 2020-11-11
Create ReadTheDocs with MkDocs

hfomar-nafcs:
I Know that ReadTheDocs needs to be rebuilt every time something changes but here is a current example built with MkDocs


hfomar-nafcs:
oh yeah, I didn't say, What I did was build ReadTheDocs with MkDocs and upload it to /docs so you can host it on GitHub Pages

hfomar-nafcs:
I have a better solution I Will Create a Request for Promptly

PR #2307, opened 2020-11-12
Created Web page with Docsify

hfomar-nafcs:
I Can't provide a Demo because the files are using the awesome-selfhosted as their root so it wont work unless the files are in your repo but you can check out their website to get an idea [docsify.js.org](https://docsify.js.org/#/) and since it's not using local md files it's pulling them from the repo and serving them when the site loads It is always up to date and does not need to be changed!!!


nodiscc:
Interesting but we could already do the same thing using https://www.sphinx-doc.org/en/master/ for generation and https://readthedocs.org/ for pages hosting, without any major kind of change to the current files (just add config files for RTD and Sphinx to the repo). Don't see how we can use your PR to provide a static site.

If you are interested in working on a HTML version of the list please check https://github.com/awesome-selfhosted/awesome-selfhosted/issues/1038, https://n8225.github.io/awesome-selfhosted-gen/, https://github.com/n8225/awesome-selfhosted-gen and https://github.com/n8225/awesome-selfhosted-gen/issues. In short, the list content must be stored as machine readable files (JSON or YAML) and a CI pipeline must take care of converting these files to a markdown/HTML document with filtering/search capabilities

Thanks anyway

hfomar-nafcs:
But The Difference is This is Rendered Directly in the browser with JS so there is no need to regenerate the files for readthedocs because it is rendering the REAMDE in the Browser!

nodiscc:
Markdown/README.md will not be the only target format (the goal is to be able to split up/filter the list by tags... one of the problems with the current format is that the list is quite large and hard to read/search) so I don't think we should include things only relevant to the current Markdown format.

The markdown file will be one of the generated formats. Docsify could be interesting to render/self-host it, we'll see to that later. [turndown](https://github.com/domchristie/turndown) could also be used. But realistically, we'll also use the static site generator generate HTML from the full README.md format (and also host a copy on Github)

PR #2489, opened 2021-04-07
Updated urls for pasty pastbin

pagdot:
Demo url wasn't working and Github url was changed (but worked because Github forwards it)


nodiscc (approved):
thanks

lus:
Thanks, @pagdot !
I forgot to do that ^^

PR #2500, opened 2021-04-15
Update README.md

mahdi1234:
Added phonetrack - https://gitlab.com/eneiluj/phonetrack-oc


nodiscc:
Sorry, we don't list apps/extensions. The nextcloud entry already points to https://apps.nextcloud.com/ and phonetrack is listed there

PR #2580, opened 2021-06-10
Remove email

Pernat1y:
Why the f**k do you even keep emails in plain text like this?!

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -348,7 +348,7 @@
 |2|Patrik Ragnarsson <patrik@starkast.net>|
 |2|Pavel Korotkiy <outdead@mail.ru>|
 |2|Pavel Lobashov <ShockwaveNN@gmail.com>|
-|2|Pernat1y <mirnesen@gmail.com>|
+|2|Pernat1y <-@gmail.com>|
 |2|Peter Demin <poslano@gmail.com>|
 |2|Peter Ivanov <peter@microweber.com>|
 |2|Phil <phil@sapphyrus.xyz>|


jnwatts:
This change won't solve your problem, see
https://github.com/awesome-selfhosted/awesome-selfhosted/blob/6fd222b250a4ea20813f69d9dcc1cdf7e571ab72/Makefile#L36

Pernat1y:
Thanks to pointing it out.
Should I update the script to 'grep -v' myself out of that list or what?

nodiscc:
see https://github.com/awesome-selfhosted/awesome-selfhosted/pull/2580

> keep emails in plain text like this

you did this yourself when you sent commits with your e-mail set in `git config user.email`. Spambots will read the git log anyway

> Should I update the script to 'grep -v' myself out

you can add an entry in https://github.com/awesome-selfhosted/awesome-selfhosted/blob/master/.github/.mailmap

nodiscc:
See also https://gitlab.com/gitlab-org/gitlab/-/issues/20960 on the topic of `git log` and GDPR compliance

Pernat1y:
The thing is, if you google my email, the only results you will get are some sorts of forks/clones of this exact repository. And this is not the only repo I made commits to. Got the idea?
Sorry for the aggressive tone, but this is just bad practice.

nodiscc:
> if you google my email, the only results you will get are some sorts of forks/clones of this exact repository

You're only making it worse...

image

> this is not the only repo I made commits to. Got the idea?

Well I understand it's a personal SEO problem. I contribute to other projects as well and this is the one that comes up in google results. Either contact google to request removal of those search results, or update https://github.com/awesome-selfhosted/awesome-selfhosted/blob/master/.github/.mailmap as I suggested.

> but this is just bad practice.

Maybe you shouldn't commit using an email address you don't want published?

pengc99:
here is my git anonymized address: pengc99@users.noreply.github.com, i just updated my git preferences and cli to use this address.

PR #2587, opened 2021-06-17
Remove Feedbin

ardichoke:
Feedbin is no longer awesome, it has moved to a paid subscription service and is explicitly directing people looking to self-host toward other solutions. Per their github page:

```Feedbin's goal is to be a great web-based RSS service. This goal is at odds with being a great self-hosted RSS reader. There are a lot of moving parts and things to configure, so for that reason I do not recommend that you run Feedbin in production.```

### DIFF
--- a/README.md
+++ b/README.md
@@ -675,7 +675,6 @@ _See https://github.com/n1trux/awesome-sysadmin#identity-management_
 **[`^        back to top        ^`](#)**
 
 - [CommaFeed](https://www.commafeed.com/) - Google Reader inspired self-hosted RSS reader. ([Source Code](https://github.com/Athou/commafeed)) `Apache-2.0` `Java`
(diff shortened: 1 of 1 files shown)


nodiscc (commented):
The website https://feedbin.com says nothing about the self-hosted service, true. But https://github.com/feedbin/feedbin looks maintained and there are installation instructions. Should we just remove the main link and use the github repo link instead?

ardichoke:
My personal opinion would be that it no longer belongs on an awesome selfhosted list since the maintainers are actively discouraging self-hosting of the software, but that's just my opinion.

nodiscc:
After re-reading their README

> I do not recommend that you run Feedbin in production.

Well if the author does not recommend it, I don't think it should be recommended in the list either.

nodiscc (approved):

Issue #1852, opened 2019-10-23
Continuous Integration checks (travis/gitlab-CI/drone/...)


n8225:
@nodiscc Checks need to be set up after move to org.  Do we want to stick with Travis, move to github actions, or one of the other CI/CD providers.  I've been using github actions on the static site https://github.com/n8225/awesome-selfhosted-gen/actions.  https://n8225.github.io/awesome-selfhosted-gen/ now updates daily.

evaryont:
We may want to consider an alternative if we want to seriously consider self-hosting this project as well. (See https://github.com/awesome-selfhosted/awesome-selfhosted/issues/89#issuecomment-526728253 and surrounding comments for that conversation.) Though, the best-in-class self-hosted CI is Gitlab's IMO but that's a hefty lift considering it doesn't integrate with anything other than Gitlab.

SuperSandro2000:
Drone + Gitea?

nodiscc:
> stick with Travis

We already have a travis config so it makes sense to go with this for the time being. I will set it up this weekend.

> move to github actions

Tying our work with proprietary, non self-hostable software even more is not the way to go IMHO. I understand it's convenient.

> best-in-class self-hosted CI is Gitlab's IMO but that's a hefty lift considering it doesn't integrate with anything other than Gitlab

True. I have been using it for some time (both SaaS and self-hosted) and it works great and is slowly becoming industry standard. I will set it up for our current [mirror](https://gitlab.com/awesome-selfhosted/awesome-selfhosted) and the repo can contain both `.travis.yml` and `.gitlab-ci.yml` without problems. We'll see if it's possible to integrate with our Github repo somehow.

> Drone + Gitea?

I have a personal Gitea instance but did not have time to set up drone yet. @SuperSandro2000 if you have a (public) drone instance we can interact with this would be a great addition, and it gives us another option.

Any suggestions are welcome.

SuperSandro2000:
> I have a personal Gitea instance but did not have time to set up drone yet. @SuperSandro2000 if you have a (public) drone instance we can interact with this would be a great addition, and it gives us another option.

no, it was on my todo to try out later but I didn't get to it yet and wouldn#t count on it. Way to much other stuff to do currently.

nodiscc:
[...]

PR #2592, opened 2021-06-20
Add github actions workflows

n8225:
Added monthly checks workflow, with a schedule to run on the 1st of the month, can also be manually ran. On completion a new issue will be created with a link to the job.  Right now the output of check-github-commit-dates.py is the only thing being posted on the issue.  Will need to update test.js to output markdown as well as a script to do that for awesome_bot. Results are also uploaded as artifacts.

I think we should also look into alternatives to awesome_bot.  https://github.com/lycheeverse/lychee looks promising, (fast, ability to use regex to prevent links checked in the python script).

Added test README.md for testing to prevent API rate limiting.

Added checks on PR.

Rewrote part of check-github-commit-dates.py, it now calls the GitHub graphql API directly. It doesn't hit the rate limit. This could use some more error checking.

@nodiscc it looks like actions isn't enabled for the repo.

#1852


.github/workflows/monthly.yml
```diff
@@ -0,0 +1,110 @@
...
+        uses: actions/setup-node@v1
+        with:
+          node-version: '14.x'
+      - name: Setup Checks
+        run: |
+          cd tests 
+          npm install --silent chalk
+          cd ..
+      - name: Set env
+        run: echo "FILE=${{ github.event.inputs.file || 'README.md' }}" >> $GITHUB_ENV;
```
nodiscc, line 51: should always be `echo "FILE='README.md' >> $GITHUB_ENV; ` but I think this is redundant anyway? Just use `github.event.inputs.file` in the later steps?

.github/workflows/monthly.yml
```diff
@@ -0,0 +1,110 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Ruby 2.6
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: 2.6.7
+      - name: Setup Checks
+        run: gem install awesome_bot
+      - name: Set env
+        run: echo "FILE=${{ github.event.inputs.file || 'README.md' }}" >> $GITHUB_ENV
```
nodiscc, line 72: same as above

.github/workflows/monthly.yml
```diff
@@ -0,0 +1,110 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup Python 3.x
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.x'
+      - name: Setup Checks
+        run: pip3 install Requests
+      - name: Set env
+        run: echo "FILE=${{ github.event.inputs.file || 'README.md' }}" >> $GITHUB_ENV
```
nodiscc, line 98: same as above

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,35 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Ruby 2.6
+        uses: actions/setup-ruby@v1
+        with:
+          ruby-version: 2.6.x
+      - name: Checks
+        run: |
+          gem install awesome_bot
```
nodiscc, line 34: ok to move to lychee but low priority, work on https://github.com/awesome-selfhosted/awesome-selfhosted/issues?q=is%3Aissue+is%3Aopen+label%3Atools will allow checks to run directly on the machine-readable data

nodiscc (changes requested):
Actions are enabled on the project but they will not trigger before the merge to master (had the same problem in https://github.com/shaarli/Shaarli/pull/1759, looks like a decent security precaution)

.github/workflows/monthly.yml
```diff
@@ -0,0 +1,110 @@
+name: Monthly Checks
+
+on:
+  schedule:
+    - cron:  '* * 1 * *'
+  workflow_dispatch:
+    inputs:
+      file:
+        description: 'File to test'
+        default: tests/README-test.md
```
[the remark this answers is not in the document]
n8225: Had this as the default for testing, so I wouldn't hit the rate limit on multiple runs.

n8225:
* So lots of changes since last commit.  Github actions API key only allows 1000 calls from the graphql api, we were close to exceeding that. So I rewrote the check-github-commit-dates.py script to perform 100 checks per call, which has the same cost as performing 1 check(surprisingly 1000 checks also had an API cost of 1).
* Rewrote test.js to output markdown.
* Using jq to create markdown from awesome_bot json output

Example issue created: https://github.com/n8225/awesome-selfhosted/issues/39
Example workflow run

n8225:
Updated examples:
* Pull request: https://github.com/n8225/awesome-selfhosted/pull/3
* Monthly checks: https://github.com/n8225/awesome-selfhosted/runs/2984733620?check_suite_focus=true
* Monthly issue: https://github.com/n8225/awesome-selfhosted/runs/2984777092?check_suite_focus=true

@nodiscc if there are any other changes you'd like let me know.

DiqahLer (approved):

Jhonnyjoy (approved):
Hi

nodiscc:
Closing, see https://github.com/awesome-selfhosted/awesome-selfhosted/issues/1852#issuecomment-1474262337

PR #2593, opened 2021-06-20
Update pasty pastebin

lus:
Hey ^^.

I updated the entry of my pastebin `pasty` and changed the following:

* I changed the position of the entry so the pastebin list is ordered alphabetically again
* I updated the demo URL (`paste.pelkum.dev` -> `pasty.lus.pm`)


nodiscc (approved):
thanks

PR #2608, opened 2021-07-04
Fix category headings

n8225:
* fix/standardize headings (can't have and H4 without an H3.)
* fix ydl_api syntax error.
* fix zero-k syntax error.


arylatt (approved):

nodiscc (approved):
thank you

Issue #2616, opened 2021-07-09
Monthly Checks - July 2021


n8225:
--------------------

### Awesome_Bot link checks

* [x] Line 446: http://fluxbb.org/, Failed to open TCP connection to fluxbb.org:80 (Connection refused - connect(2) for "fluxbb.org" port 80)
* [ ] Line 450: https://gosora-project.com/, 
* [ ] Line 463: https://nitter.net, SSL_connect returned=1 errno=0 state=error: certificate verify failed (unable to get local issuer certificate)
* [x] Line 655: https://pyshelf.com, Failed to open TCP connection to pyshelf.com:443 (Connection refused - connect(2) for "pyshelf.com" port 443)
* [ ] Line 803: https://v2.femto.pw/, Failed to open TCP connection to v2.femto.pw:443 (Connection refused - connect(2) for "v2.femto.pw" port 443)
* [x] Line 1007: https://demo.funkwhale.audio/, 
* [x] Line 1177: https://domchristie.github.io/turndown/, 
* [x] Line 1291: https://maxvoltar.photo/, Failed to open TCP connection to maxvoltar.photo:443 (getaddrinfo: No address associated with hostname)
* [ ] Line 1399: https://usegood.work/, Failed to open TCP connection to usegood.work:443 (Connection timed out - connect(2) for "usegood.work" port 443)
* [x] Line 171: https://github.com/Totonyus/ydl_api, Dupe

--------------------
#### There were 185 repos with issues.
* [ ] 2013-06-17, https://github.com/thibaud-rohmer/PhotoLight 
* [ ] 2015-01-24, https://github.com/mil/markdown-tree 
* [ ] 2015-12-01, https://github.com/georgeOsdDev/markdown-edit 
* [ ] 2016-05-10, https://github.com/lgg/tasks.php, Moved to https://github.com/lgg-archive/tasks.php 
* [ ] 2016-09-17, https://github.com/nhoening/gritttt-rss 
* [ ] 2016-09-19, https://github.com/piero-la-lune/Bumpy-Booby 
* [ ] 2016-10-23, https://github.com/benschwarz/gallery-css 
* [ ] 2016-10-23, https://github.com/matrix-org/matrix-angular-sdk 
* [x] 2016-11-17, https://github.com/n1try/http2-serverpush-proxy, Archived, Moved to https://github.com/muety/http2-serverpush-proxy 
* [ ] 2017-04-04, https://github.com/haldean/docstore 
[...]

PR #2617, opened 2021-07-09
Fixes and Updates for July monthly checks

n8225:
Updates/Fixes for Monthly Checks - July 2021 #2616 

This can be squashed if needed, I think the git log is cleaner this way though.


nodiscc (approved):
Great, thank you

PR #2637, opened 2021-08-04
remove table of contents, use github's auto-generated TOC

nodiscc:
image


DanielRamp:
Is there a way to show the **previous "in file" toc**?
It was way more clear.

+ had more browser height
+ was already open/there
- ok, yes you had to scroll over to get to the content, but thats all, right?

samyak-jain:
Yeah. It makes sense why you would do this. But, I do think this is much more confusing mostly because Github's TOC UX is really bad. It took me a while to figure out where the TOC was. Maybe at least posting the screenshot from the PR in the Readme for the time being should help?

nodiscc:
We could make it

> Click on the menu icon next to  image README.md for a list of sections

(add the icon to make it less confusing).

The underlying problem is that Github [does not have a proper automatic TOC](https://stackoverflow.com/questions/9721944/automatic-toc-in-github-flavoured-markdown) ([Gitlab does](https://docs.gitlab.com/ee/user/markdown.html#table-of-contents)).

If you want to send a pull request with an up-to-date table of contents, please do. But I cannot guarantee that it will stay up-to-date (though sections should not change much anymore).

DanielRamp:
> If you want to send a pull request with an up-to-date table of contents, please do. But I cannot guarantee that it will stay up-to-date (though sections should not change much anymore).

What about a automatic generation? like [this](https://github.com/ekalinin/github-markdown-toc#toc-generation-with-github-actions) or as alternative I [found](https://github.com/ycd/toc) [these](https://github.com/thlorenz/doctoc) [three](https://github.com/kubernetes-sigs/mdtoc) there is a [github topic](https://github.com/topics/table-of-contents).

I might try to do a pr...

hashborgir:
I just posted about this too. Ridiculous. When you know Github's UI is not good, don't use it. Or use it and still provide a ToC in the document. How often are you planning on changing the section headings/names/anchors?

PR #2638, opened 2021-08-04
sort section contents alphabetically

nodiscc:
Most sections were no longer sorted correctly


nodiscc:
This also unindents all sub-items

PR #2639, opened 2021-08-06
Update Standard Notes domain

mobitar:
We've moved to .com :)


slel:
But the demo link stays at .org?

mobitar:
Yup for now. We're doing the migration in parts :)

slel:
The page https://github.com/standardnotes
has a header with the new .com home url.
The docs url there could be updated too
since it already redirects from .org to .com.
And the help email if/when it has migrated.

mobitar:
Docs domain updated 👍

Help email currently can receive at help@standardnotes.com. We're currently in the process of making that the primary email domain as well.

nodiscc (approved):

PR #2662, opened 2021-09-01
auto generated table of content (implementation)

DanielRamp:
- implemented automated generation of table of content
- added toc marker to README.md
- removeed the message about the built-in toc of github

### DIFF
--- a/.github/workflows/auto_table_of_content.yml
+++ b/.github/workflows/auto_table_of_content.yml
@@ -1,0 +1,18 @@
+on:
+  push:
+    branches: [master]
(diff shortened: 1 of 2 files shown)


nodiscc:
Thank you but the TOC generation should be done with https://github.com/nodiscc/hecat which is what we're planning to use to generate the list. The current problem is that ToC generation is not implemented yet in the python script (https://github.com/nodiscc/hecat/runs/3243642587). https://github.com/nodiscc/hecat/issues/5. Patches welcome

PR #2663, opened 2021-09-02
Removed demo link from Mailu entry

### LINKED ISSUE #2645: Mailu Demo link leads to an empty wiki
The demo for the Mailu application gets you to https://github.com/Mailu/Mailu/wiki/Demo-server and that page does not exist, thus allowing the users to create a new page. 

We would most likely want to remove the demo link from in front of the Mailu entry.

t0xic0der:
The demo link leads to an empty wiki which is publicly editable

Signed-off-by: Akashdeep Dhar <akashdeep.dhar@gmail.com>

Fixes #2645


nodiscc (approved):
Thanks

PR #2727, opened 2021-10-20
Workflow to run syntax test on pull requests

josefandersson:
Maybe the plan is to use the `github_actions` branch in the future, but it seems like a waste not to run the syntax test on all pr's right now.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -1,0 +1,14 @@
+name: Run syntax test
+
+on:
+  pull_request:
+    branches: [ master ]
+  workflow_dispatch:
+
+jobs:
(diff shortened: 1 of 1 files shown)


nodiscc:
> Maybe the plan is to use the github_actions branch in the future

No, the plan is https://github.com/awesome-selfhosted/awesome-selfhosted/issues/1038

> it seems like a waste not to run the syntax test on all pr's right now.

No, the makefile-based syntax check is unmaintained and will be removed. I don't know if it still works correctly. Please check https://github.com/awesome-selfhosted/awesome-selfhosted/issues/1038 and the 1.0 milestone https://github.com/awesome-selfhosted/awesome-selfhosted/milestone/2

Issue #2772, opened 2021-12-07
crtmpserver Broken Link


F0x1fy:
It appears crtmpserver has a broken link and may not longer be available on Debian? I was able to find [a clone of the SVN repository](https://github.com/j0sh/crtmpserver) but it's unofficial and seems out of date. The official crtmpserver website also seems to be under a new owner completely unrelated to crtmpserver. This project just seems overall abandoned. Possible removal?

nodiscc:
yep, it should be removed. Can you send a pull request?

F0x1fy:
Hey, sorry. I’ve had a lot come up in my personal life and I completely forgot about this. I’ll submit a PR either today or tomorrow

PR #2792, opened 2021-12-21
Removed crtmpserver (abandoned project)

F0x1fy:
crtmpserver has a broken link and is no longer available on Debian. Only remains of this project are old forks, and website is under a new owner. See issue #2772


nodiscc (approved):
thanks

PR #3027, opened 2022-05-24
Update information on Bolt CMS

henk23:
The repo for Bolt CMS moved for version 5, and the old domain's certificate has expired as of today. Also, I cannot find an interactive demo anymore.


tdung (approved):

nodiscc (approved):
thanks

tdung (commented):
Y

PR #3352, opened 2022-10-10
Remove dead link

ndsvw:
Remove dead link to http://cubny.com/lite-reader/


nodiscc (approved):
Demo site is down and the last commit is from 2016. Uses PHP 5.3

Thanks

PR #3666, opened 2023-03-20
Added an external link - Search across all selfhosted repos.

rasulkireev:


rasulkireev:
Done, thanks for the note!

nodiscc:
I don't think this is very ready/usable yet:
- It doesn't list all projects, only those whose source code is hosted on Github
- It doesn't display licenses
- It doesn't display correct languages/platforms (in the sense "The Language tag is the main server-side requirement for the software")
- It doesn't display categories, the user-provided tags are not very useful to users (`hacktoberfest`, `encryption`...)
- It only displays a few items by default
- It doesn't display links to project homepages, only to source code repositories
- ...

By the way your typesense API key is publicly accessible (`BIZKrznddXvifP7KEXAZn3pItniYP8nJ`), I don't know if this is on purpose or could be abused.

Anyway, I don't think this is an improvement over using `Ctrl+F` on the original markdown list, using our table of contents, or using Github search directly. Hence closing.

rasulkireev:
@nodiscc , Thank you for the comments on my site. I will try to improve on those, adding usability and interesting spin on the search.

Issue #3558, opened 2023-02-01
Remove unmaintained projects, dead links


nodiscc:
This is a list of unmaintained software projects and dead links in awesome-selfhosted.

Feel free to help fixing these issues by following the usual [contribution](https://github.com/awesome-selfhosted/awesome-selfhosted/blob/master/.github/PULL_REQUEST_TEMPLATE.md) process (edit the README.md file, propose your change through a Pull Request).

> Software with no development activity for 6-12 months may be removed from the list.

Remember that some projects are feature-complete/mature/simple enough that they don't need much maintenance. Check if they still have active users/work as advertised before removing them. You may notify the maintainer of each project in comments of your Pull Request, to inquire about the status of their project.

A summary of current problems can be found below. Please refer to the results of the latest daily automated checks for more information: [![](https://github.com/awesome-selfhosted/awesome-selfhosted-data/actions/workflows/check-dead-links.yml/badge.svg)](https://github.com/awesome-selfhosted/awesome-selfhosted-data/actions/workflows/check-dead-links.yml) [![](https://github.com/awesome-selfhosted/awesome-selfhosted-data/actions/workflows/check-unmaintained-projects.yml/badge.svg)](https://github.com/awesome-selfhosted/awesome-selfhosted-data/actions/workflows/check-unmaintained-projects.yml)

-----------------------

Projects without commits for > 365 days:

```
ERROR:awesome_lint.py: Actionsflow: last updated -382 days, 7:57:09.478037 ago, older than 186 days
ERROR:awesome_lint.py: Adminer: last updated -822 days, 7:57:09.477427 ago, older than 186 days
ERROR:awesome_lint.py: AmIUnique: last updated -1620 days, 7:57:09.476853 ago, older than 186 days
ERROR:awesome_lint.py: asciiflow: last updated -561 days, 7:57:09.475457 ago, older than 186 days
ERROR:awesome_lint.py: Bicimon: last updated -471 days, 7:57:09.473635 ago, older than 186 days
ERROR:awesome_lint.py: Chowdown.io: last updated -528 days, 7:57:09.470078 ago, older than 186 days
ERROR:awesome_lint.py: Cubiks-2048: last updated -3340 days, 7:57:09.467891 ago, older than 186 days
ERROR:awesome_lint.py: CUPS: last updated -398 days, 7:57:09.467739 ago, older than 186 days
[...]

PR #3870, opened 2023-06-04
Fix papermerge url

Rabenherz112:
They moved their website from www. to their main domain.

ref: #3558
`https://www.papermerge.com : HTTPSConnectionPool(host='www.papermerge.com', port=443): Max retries exceeded with url: / (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate (_ssl.c:997)')))`


nodiscc (approved):
Thank you

PR #3874, opened 2023-06-05
Fix Nextcloud Memories

Rabenherz112:
This PR aims to resolve the issue with the non-functional demo link. It appears that the demo has been relocated to their own domain. Therefore, I have taken the initiative to include the main website in this entry as well.

ref: #3558
`https://memories-demo.radialapps.com/ : HTTP 526`


nodiscc (approved):
Thanks

PR #3900, opened 2023-06-14
Remove Demo from LiveKit

Rabenherz112:
The demo seems to get removed from the Website.

ref: #3558 
`https://livekit.io/playground : HTTP 404`


Rabenherz112:
Ahh, sorry.. Just seen duplicate, since nodiscc already has a PR open for that..

PR #3905, opened 2023-06-17
Remove Tasks from Task Management & To-do Lists

Rabenherz112:
The purpose of this PR is to remove Tasks since it appears to be unmaintained. In the Install Instructions, it still states 'PHP 5.5.9 or higher', which is already long EOL. Installation with newer (supported) versions of PHP seems to result in errors (https://github.com/m1guelpf/Tasks/issues/7).

ref: #3558 
`WARNING:awesome_lint.py: Tasks: last updated -1478 days, 23:22:41.638636 ago, older than 365 days`


nodiscc (approved):
Thanks

PR #3908, opened 2023-06-18
Remove BicBucStriim from Document Management - E-books

Rabenherz112:
The purpose of this PR is to remove BicBucStriim from the List due to its apparent lack of maintenance and absence of installation instructions.

ref: #3558 
`WARNING:awesome_lint.py: BicBucStriim: last updated -429 days, 23:22:41.654973 ago, older than 365 days`


nodiscc (approved):
Thanks

PR #3923, opened 2023-06-22
Remove Photonix (unmaintained since 2022)

Rabenherz112:
It appears that Photonix has not received any maintenance since 2022. Another concern regarding the project's activity is currently raised in this GitHub issue: https://github.com/photonixapp/photonix/issues/426. Additionally, it is worth mentioning that Dependabot updates have been halted due to the project's inactivity. Presently, there are still 19 pending dependency updates.

ref: #3558 
`WARNING:awesome_lint.py: Photonix: last updated -456 days, 23:18:32.066054 ago, older than 365 days`


nodiscc (approved):
Thanks

PR #3926, opened 2023-06-23
Remove ImageStore (unmaintained since 2022)

Rabenherz112:
The projects appear neglected, with 31 pending dependency updates and an inactive Dependabot due to the lack of project activity. Furthermore, it seems that the creator is also inactive on GitHub.

ref: #3558 
`ImageStore: last updated -435 days, 23:16:59.907264 ago, older than 365 days`


Rabenherz112:
@gregordr is https://github.com/gregordr/ImageStore still being actively maintained?

nodiscc (approved):
Thanks

PR #3952, opened 2023-06-29
Remove Paste (unmaintained since 2021)

nodiscc:
- ref. #3558
- `WARNING:awesome_lint.py: Paste: last updated -555 days, 4:54:42.041576 ago, older than 365 days`
- Not compatible with PHP8 https://github.com/jordansamuel/PASTE/issues/202


jordansamuel:
huh

PR #3945, opened 2023-06-29
Update language tags

Rabenherz112:
I have noticed that some of the software I use does not include all the languages in their installation options (especially Docker or Kubernete). This can be quite frustrating at times because I may install a particular software only to later discover that there is a better alternative available, but I didn't consider it initially because it was not compatible with my environment.

Therefore, the purpose of this pull request is to enhance the language support of the software by adding the languages that are currently supported along with their respective installation instructions. While I may still miss some programs or languages, this effort should significantly improve the situation.

Please note that this pr is currently a draft, and I will mark it as ready once I have completed the list. However, I wanted to provide an update on the progress and be available for any discussions or address any concerns that may arise.

The fact that the table of contents was edited is a bug in VS Code (it does this for some reason with every commit), and once I'm done, the whole thing will be reverted. (https://github.com/awesome-selfhosted/awesome-selfhosted/commit/ad658c943890b19230012653fcac8db540bf9986)

Edit:
Note for me, next up Communication - IRC

PR #3976, opened 2023-07-06
Update language tags - Part 2

Rabenherz112:
ref: #3945 

Edit: Next up: Feed Readers


Rabenherz112:
Could someone kindly review "Tokumei"?  I'm a bit puzzled because the language mentioned is 'rc', which I don't know what this is. Upon examining the installer, it appears to be more like Go? Currently, I'm a bit confused their..

nodiscc:
> Could someone kindly review "Tokumei"? I'm a bit puzzled because the language mentioned is 'rc', which I don't know what this is. Upon examining the installer, it appears to be more like Go? Currently, I'm a bit confused their..

The project README reads:

> Tokumei is written in rc, the Plan 9 shell, using the werc minimal web framework

**https://en.wikipedia.org/wiki/Rc**

The main installation script can be found at https://tokumei.co/hosting/private-clearnet -> https://tokumei.co/privclear.sh. It looks designed for Debian-based systems and uses `apt` to install the [9base](https://packages.debian.org/bookworm/9base) packkages which contains the `rc` interpreter. It clones Tokumei sources (definitely written in `rc` shell https://gitlab.com/tokumei/tokumei/-/tree/master/bin), installs https://github.com/uriel/cgd which is a FastCGI server that serves exposes the `rc` shell application as a HTTP service on local port 3333, installs an nginx configuration that redirects HTTP requests to cgd. `cgd`  is written in go but is only used as a middleware, not the main application. So `rc` is correct.

But anyway this project has been unmaintained since 2017 (https://gitlab.com/tokumei/tokumei/-/commits/master?ref_type=heads), and cgd is unmaintained since 2012. I think it's safe to remove it.

Rabenherz112:
I found "VanDAM" in the "Document Management - E-books" section.
As it describes itself, it seems an asset/file manager with nothing related to E-Books:
> VanDAM is a Digital Asset Manager (DAM), specifically designed for 3D print files

Should we move it from there to Miscellaneous?

nodiscc:
> VanDAM is a Digital Asset Manager (DAM), specifically designed for 3D print files
>Should we move it from there to Miscellaneous?

I think this was overlooked when it was submitted (https://github.com/awesome-selfhosted/awesome-selfhosted/pull/3652). It should definitely be moved, thank you.

nodiscc:
Great work, thank you. Can you please resolve merge conflicts?

nodiscc (approved):

PR #3984, opened 2023-07-09
Move VanDAM to Miscellaneous

Rabenherz112:
As it describes itself, it seems an asset/file manager with nothing related to E-Books:
> VanDAM is a Digital Asset Manager (DAM), specifically designed for 3D print files

ref: #3976


kokomo123 (approved):
No problems with this.

nodiscc (approved):
Thanks

PR #3998, opened 2023-07-10
Pull Request template: require forks to provide a list of differences with the original project

nodiscc:
- ref. https://github.com/awesome-selfhosted/awesome-selfhosted/pull/3467

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -24,6 +24,7 @@ To ensure your PR is dealt with swiftly please check the following:
   If not, your addition may be inserted into `Miscellaneous`.
 - [ ] Any software project you are adding to the list is actively maintained.
 - [ ] Any software project you are adding has working installation instructions.
+- [ ] If you are adding software forked from another project, please provide/link to a clear list of differences between both.
 - [ ] The pull request title is informative, unlike "Update README.md".
   Suggested titles: "Add aaa to bbb" for adding software aaa to section bbb,
   "Remove aaa from bbb" for removing, "Fix license for aaa", etc.


kokomo123 (approved):
No issues with this, it makes sense given differences and yeah, i should have realized that when trying to include Forgejo into the list.

Rabenherz112:
I also support the idea of requiring forks to provide an explanation of the main differences between their software and the original upstream software in a PR. This could become especially important when a PR is submitted for a fork where the base software is already on this list.

Additionally, I had an idea: would it be sensible to enforce the inclusion of a brief statement like `This is a fork of [XYZ](Link to base)`? Such a requirement would enable users to easily recognize when something is a fork of another project.

Rabenherz112 (approved):

nodiscc:
> would it be sensible to enforce the inclusion of a brief statement like This is a fork of [XYZ](Link to base)? 

This can be requested during reviews on a case-by-case basis, but I don't think it's necessary to add it to the pull request template (it's already quite long).

As far is know there have been no attempts to disguise a fork as an original work so far, so a written rule might not be in order - but you can open open a new issue or PR if you want to discuss this further.

Also inline links are discouraged because, in my opinion, they make the list harder to read (and I think they used to confuse the dead link/syntax checker but that may no longer be the case). So I'd prefer simply adding `(fork of XYZ)`- however, searching for `fork` in the list you can see that there are inconsistencies, so maybe a bit of cleanup is in order to standardize these:

- Radarr is an independent fork of Sonarr
- forked from Radicale
- actively developed fork of RainLoop
- Fork of Prosody IM
- A fork of paperless,
- (fork of Newspipe) _<- in my opinion this should be the standard/accepted syntax_
- Fork of Snapdrop.
- (fork of [Airtime](https://github.com/sourcefabric/Airtime)).
- It is a a fork of Overseerr
- Community managed fork of Gogs
- Forked from Crepido.

Rabenherz112:
It was just an idea, so yeah. 
When I have some spare time, I'll submit a pull request to ensure that all fork mentions are consistently formatted as "(fork of XYZ)." (Also a fan of this formatting).
After giving it further consideration, I concur that including a hyperlink might lead to confusion regarding the actual software (if users click on it, assuming it's the software listed) since hyperlinks are primarily intended for the software's own pages. Additionally, if it could potentially raise issues with the link checker, this makes the formatting anyway a no-go.

PR #4062, opened 2023-07-28
Remove golinks (unmaintained since June 2022)

nodiscc:
- https://git.mills.io/prologic/golinks/commits/branch/master
- last updated in 2021 before that
- ref. #3558


KieranRobson (approved):

Shednandezstore:
Okay

PR #4067, opened 2023-07-30
fix: Typos and Dead Link

KieranRobson:
Couple of typos and dead link removed. Couldnt find an alternative to replace it with.


nodiscc (approved):
Thank you

PR #4192, opened 2025-03-31
fix: issue template to new github format

Rabenherz112:
Lately, we've started receiving issues again in the old repository. To prevent this, we should fix the currently broken issue template by adding an automatic redirect to the new repository. This way, issues can no longer be opened in the current repo.

Preview of the change:

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,5 @@
+blank_issues_enabled: false
+contact_links:
+  - name: "Awesome Selfhosted Data"
+    url: "https://github.com/awesome-selfhosted/awesome-selfhosted-data/issues"
+    about: "Please submit your issues or suggestions to the awesome-selfhosted-data repository instead."
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,1 +1,0 @@
-Please do not open issues in this repository. Use https://github.com/awesome-selfhosted/awesome-selfhosted-data instead.


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,5 @@
+blank_issues_enabled: false
+contact_links:
+  - name: "Awesome Selfhosted Data"
+    url: "https://github.com/awesome-selfhosted/awesome-selfhosted-data/issues"
```
TheGiraffe3, line 4: The other option would be to change this to the choose page, which makes it more clear. Not entirely necessary though.
```suggestion
    url: "https://github.com/awesome-selfhosted/awesome-selfhosted-data/issues/new/choose"
```

.github/issue_template.md
TheGiraffe3, line 1: Better to delete this file instead of removing its contents.

Rabenherz112: Wdym?
https://fs.ravshort.com/89ZKe.png

.github/ISSUE_TEMPLATE/config.yml
Rabenherz112: I thought about this as well, but only leading them to the new issue page, means they can look through them first before creating one, which might reduce possible duplicates.

.github/issue_template.md
Ki-er: Can confirm file deletion.

TheGiraffe3: Odd, hadn't shown as deleted for me before. This can be resolved.

nodiscc:
This is because I had previously enabled interaction limits on the old repository, but the limitation period expired (it can only be limited up to 6 months, and there is no notification when it expires).

I just re-enabled it for another 6 months

image


Given this, do you think this PR is still needed.

Rabenherz112:
I think redirecting them anyway is still a better solution. If they want to open an Issue they still can, but directly at the correct repo. Don't see any disadvantage of this.

nodiscc (approved):

